PR #29, opened 2015-09-11
Fix paths when adding dependencies to plt closes #25

### LINKED ISSUE #25: `plt_add_deps: true` option fails
Thanks for sharing this! I'm just getting started and tried the configuration `dialyzer: [plt_add_deps: true]`.

The first issue I ran into is that this seems to have no effect if updating an existing PLT. So I deleted the file to regenerate a new one.

The second issue is the command output from `mix dialyzer.plt` appears to concatenate all -pa options into one string:
```
dialyzer --output_plt /Users/JC/.dialyxir_core_17_1.0.5.plt --build_plt -padeps/phoenix/ebin-padeps/phoenix_html/ebin-padeps/phoenix_ecto/ebin-padeps/postgrex/ebin-padeps/phoenix_live_reload/ebin-padeps/cowboy/ebin --apps erts kernel stdlib crypto public_key phoenix phoenix_html phoenix_ecto postgrex phoenix_live_reload cowboy  -r /usr/local/Cellar/elixir/1.0.5/bin/../lib/elixir/../eex/ebin /usr/local/Cellar/elixir/1.0.5/bin/../lib/elixir/../elixir/ebin /usr/local/Cellar/elixir/1.0.5/bin/../lib/elixir/../ex_unit/ebin /usr/local/Cellar/elixir/1.0.5/bin/../lib/elixir/../iex/ebin /usr/local/Cellar/elixir/1.0.5/bin/../lib/elixir/../mix/ebin
```
(As an aside the correct paths are actually _build/dev/lib/..., so an option to the root deps path might make sense as well).

Am I doing something wrong or is this a bug? I'm on OSX 10.10.4, Elixir 1.0.5

tomekowal:


jeremyjh:
Awesome, thanks!

Issue #31, opened 2015-10-15
dialyxir forgets paths for mix applications on plt update


grundrauschen:
Hello,

I think I found a bug in dialyxir.
When I install dialyxir as mentioned and don't create the plt inside the repository, but inside a software I want to test, which has the attribute `plt_add_apps: true` set, it builds the plt with including all dependencies and its paths in `_build/dev/lib` just fine and works well.
When I instead create a plt in dialyxir with `mix dialyzer.plt` first and then try to redo this command in the project I want to test, i get the following result

```
$ mix dialyzer.plt
Some apps are missing and will be added:
phoenix ecto phoenix_ecto mariaex ...
Adding Erlang/OTP Apps to existing PLT ... this will take a little time
dialyzer --add_to_plt --plt /Users/tobias/.dialyxir_core_18_1.1.1.plt --apps phoenix ecto phoenix_ecto mariaex ...

dialyzer: No such file, directory or application: "phoenix"

mix dialyzer.plt  14.72s user 0.48s system 100% cpu 15.140 total
```
So it seems, that when dialyxir is adding modules to the plt instead of creating it initially, it forgets all `-pa` parameters.

Cheers,
Tobias

PR #32, opened 2015-11-11
Fix missing -pe option

ishikawa:
I ran into the same problem reported in #31. Under the project which have `plt_add_deps: true` option, `mix dialyxir.plt` doesn't add `-pe` option if PLT file already exists.

```
% mix dialyzer.plt
Some apps are missing and will be added:
poison httpoison joken phoenix phoenix_ecto mariaex phoenix_html phoenix_live_reload cowboy
Adding Erlang/OTP Apps to existing PLT ... this will take a little time
dialyzer --add_to_plt --plt /Users/takanori_is/.dialyxir_core_18_1.2.0-dev.plt --apps poison httpoison joken phoenix phoenix_ecto mariaex phoenix_html phoenix_live_reload cowboy

dialyzer: No such file, directory or application: "poison"
```


jeremyjh:
Thanks!

Issue #33, opened 2015-12-26
Pass dialyzer exit status on to mix


schnittchen:
It would be nice to have access to the exit status of the dialyzer run, to have the benefit of using mix for it in CI (or any scripted environment).

jeremyjh:
When you say have the exit status, you mean have Mix exit with the same status code as returned by dialyzer?

I could not find any support for this in the Mix API. 

I confirmed I can call :erlang.halt with the return code from dialyzer but this would break things like mix command chaining. Still, I do not see why this couldn't be enabled with an option you pass, e.g.

`mix dialyzer --halt-exit-status`

Would that work for you?

schnittchen:
Yes that would work. Alternatively, a mix task outputting the exact dialyzer command line would work as well (and be potentially more useful in the future).

jeremyjh:
You mean just have it print the dialyzer command to stdout without invoking it? That would be easy enough. What do you mean by parse-able fashion though?

schnittchen:
Would something like

```
DIALYZER_COMMAND=`mix dialyzer --print-command`
$DIALYZER_COMMAND
```

be possible? Might pose escaping problems, but that's the general idea.

schnittchen:
The --halt-exit-status way was easy enough :)

PR #34, opened 2016-01-01
support --halt-exit-status option

schnittchen:
This implements #33. I could not find a way for dialyzer to exit with status 1, so I could check only 0 and 2.

### DIFF
--- a/lib/mix/tasks/dialyzer.ex
+++ b/lib/mix/tasks/dialyzer.ex
@@ -7,7 +7,9 @@ defmodule Mix.Tasks.Dialyzer do
 
   ## Command line options
 
-    * `--no-compile`      - do not compile even if needed.
+    * `--no-compile`       - do not compile even if needed.
+    * `--halt-exit-status` - exit immediately with same exit status as dialyzer.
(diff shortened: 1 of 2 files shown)


jeremyjh:
Hi, Thanks! 

A suggestion, perhaps we should check if exit status equals 0, and exit normally if so. That way if you are chaining mixing commands it only halts on a dialyzer issue.

Also could you please update the @moduledoc to describe the new argument, and bump the version in mix.exs?

schnittchen:
When the new option is used, I'd expect the exit status to be exactly the same as dialyzer's. That is no longer the case when chaining is possible.

I would rather keep the code as is and document that behaviour.

jeremyjh:
Ok thats fine, I agree just point that out in the doc string.

schnittchen:
Commits added.

jeremyjh:
Thanks, if you'll just squash the commits I will merge. You can do this with a rebase -i and fixup the second two commits then push it with a -f flag.

schnittchen:
Done.

jeremyjh:
Thanks!

PR #36, opened 2016-01-11
improve support for umbrella projects

vadim-moz:
set default for `paths` for umbrella projects to list of in-umbrella
applications.


jeremyjh:
Thanks!

PR #38, opened 2016-01-19
Fix umbrella app build paths to include Mix.env in them. Fix typo.

petrikero:
My umbrella project was failing to dialyze with dialyxir, because the 'dialyzer_paths' for the projects included in the umbrella were all missing the Mix.env portion of the path. The paths used were of the form '_build/lib/app_name/ebin', when the beam files were actually in '_build/dev/lib/app_name/ebin'. Using Mix.Project.build_path() instead of Mix.Project.app_path() fixed this for me.

Are there cases where the old behavior is correct and a check would be needed whether to use the old or the new behavior?

Also fixed a typo: childeren -> children.

### DIFF
--- a/lib/mix/tasks/dialyzer.ex
+++ b/lib/mix/tasks/dialyzer.ex
@@ -57,19 +57,19 @@ defmodule Mix.Tasks.Dialyzer do
     || ["-Wunmatched_returns", "-Werror_handling", "-Wrace_conditions", "-Wunderspecs"]
   end
 
-  defp umbrella_childeren_apps do
+  defp umbrella_children_apps do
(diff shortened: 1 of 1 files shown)


jeremyjh:
I don't understand, it looks like you are just changing the name of a local function. If you have an umbrella project it is calling the app_paths which uses Mix.project.build_path already. How did renaming defp app_paths  fix your issue?

petrikero:
Uhh, you're right. I must've gotten confused with the latter call to .app_path() or something. The actual problem being that I was using the old v0.3.0 from hex.pm that didn't include the umbrella fix at all. Sorry about the confusion.

PR #40, opened 2016-02-08
Updates `plt_add_deps` option to ignore deps from other Mix envs

howleysv:


jeremyjh:
Hi, thanks for the PR. I am wondering if this change should be opt-in. Reason is it applies to the plt generation which does not have to be the same environment in which checks are run. Realistically though probably everyone would do both tasks in dev. Can you describe the issue you have that you are trying to solve with this?

howleysv:
The plt generation looks for compiled beam files under `_build/<env>/lib/<dep>/ebin` but not all deps are compiled for every env.

E.g. I have `bypass` and `excoveralls` as `:test` only deps
```
$ mix dialyzer.plt
Some apps are missing and will be added:
bypass excoveralls
Adding Erlang/OTP Apps to existing PLT ... this will take a little time
dialyzer --add_to_plt --plt ~/.dialyxir_core_18_1.2.0.plt <...> -pa _build/dev/lib/bypass/ebin -pa _build/dev/lib/excoveralls/ebin -pa _build/dev/lib/dialyxir/ebin --apps bypass excoveralls

dialyzer: Bad directory for -pa: _build/dev/lib/bypass/ebin
```

jeremyjh:
Thanks!

howleysv:
np :+1:

PR #55, opened 2016-11-08
Fix instructions to install globally

eksperimental:
Add `deps.get` otherwise this error is given.

```sh
Unchecked dependencies for environment dev:
* ex_doc (Hex package)
  the dependency is not available, run "mix deps.get"
** (Mix) Can't continue due to errors on dependencies
```


jeremyjh:
Thanks!

eksperimental:
you're welcome!

PR #59, opened 2016-12-19
Enable dialyzer.ignore-warnings

melpon:
I want to ignore well-known dialyzer warnings.

It seems that Erlang users make a file `dialyzer.ignore-warnings` and run command:

```
dialyzer ... | fgrep -v -f ./dialyzer.ignore-warnings
```

Many Erlang projects using [dialyzer.ignore-warnings](https://github.com/search?utf8=%E2%9C%93&q=filename%3Adialyzer.ignore-warnings&type=Code&ref=searchresults)

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,4 +133,48 @@ end
 
 ### Ignore Warnings
 
-**TODO**
+If you want to ignore well-known warnings, you can specify file path to `:ignore_warnings`.
+
+```elixir
+def project do
+ [ app: :my_app,
+   version: "0.0.1",
+   deps: deps,
+   dialyzer: [ignore_warnings: "dialyzer.ignore-warnings"]
+ ]
+end
+```
+
+If an each line of dialyzer output matches an any each line in `"dialyzer.ignore-warnings"`, the line is filtered.
+
+For example, dialyzer outputs here:
+
+```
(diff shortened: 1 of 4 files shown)


jeremyjh:
Hi, thanks for submitting this - I agree we need this feature.

I wonder though if writing a temp file and shelling `fgrep` is the best way - could we filter natively in Elixir instead? Maybe just split the output lines and do a filter/match on them?

melpon:
Agree. I think using `fgrep` is not good solution.
https://github.com/jeremyjh/dialyxir/pull/59/commits/cf5be0a5d768d876b963b8f58d6cfc9af213f3fb changed to use Elixir native code instead of `fgrep`.

jeremyjh:
Yes, I like this a lot better.

I think we just need to add an explanation to the README and the module doc string.

melpon:
I added documents to the README and the module doc.
However, I'm not native English speaker. If written English is broken, please fix it.

jeremyjh:
merged in e3759ac72499c05653913f9506a20c7b18496740

jeremyjh:
@melpon - Thanks for this, sorry for the delay getting it merged. This has been pushed to Hex in version 0.4.2.

Issue #62, opened 2016-12-29
Full path to files in warnings


Zert:
I have the project structure like this:
```
lib
    abcd
        protocol.ex
    efgh
        protocol.ex
    ijkl
        protocol.ex
```

Elixir allow this, it is awesome!

When I run `mix dialyzer` I see a warning:

```
...
protocol.ex:230: The pattern {'error', _error@1} can never match the type 'ok'
...
```

but I cannot say certain, which of `protocol.ex` files affected.

Could you please add the full path to files? May be from the current directory.

jeremyjh:
Currently this isn't possible because `dialyxir` just shells the Erlang `dialyzer` program and I do not have control of its output. In a future release I do plan to switch to using the API and then would be able to support this.

PR #63, opened 2017-01-04
Display full paths of offending files

joshuawscott:
This changes the output to always display the path to the file:
`lib/my_lib/foo/my_module.ex:32: Function publish/6 has no local return`
instead of
`my_module.ex:32: Function publish/6 has no local return`

I looked and didn't find any previous discussion of this. To me it doesn't seem like there is any reason to not display the full path.


jeremyjh:
Actually @Zert was just asking for this in #62 but I blew him off - I should have checked the `dialyzer` doc first. Thanks for the PR!

PR #64, opened 2017-01-05
Fix parentheses warnings for Elixir 1.4

xtian:
Elixir 1.4 gives this warning:
```
warning: variable "dialyzer_config" does not exist and is being expanded to "dialyzer_config()",
please use parentheses to remove the ambiguity or change the variable name
  lib/dialyxir/project.ex:16
```


jeremyjh:
Thanks!

PR #65, opened 2017-01-05
Plt: HashSet -> MapSet

xtian:
HashSet is deprecated and causes this warning in Elixir 1.4:

```
warning: HashSet.new/0 is deprecated, use the MapSet module instead
  lib/dialyxir/plt.ex:28
```


jeremyjh:
Thanks!

PR #66, opened 2017-01-08
Enable warnings-as-errors on CI and fix parentheses warnings in mix task

xtian:
This changes the build to pass `--warnings-as-errors` to `mix compile` so warnings don't sneak through. :)


jeremyjh:
Great idea, thanks!

PR #68, opened 2017-01-17
Add color to warnings

### LINKED ISSUE #67: Pretty output
I was wondering what your thoughts were on 'prettying' up the output from dialyzer so that the warnings would stand out a bit more. I hacked together a crude approach [here](https://github.com/daveshah/dialyxir/commit/421454c66d9dabeccd03d211fd46e1c2eac39090). Thoughts? :)

daveshah:
This commit adds simple coloring to dialyzer warnings.

Closes #67

### DIFF
--- a/lib/dialyxir/output.ex
+++ b/lib/dialyxir/output.ex
@@ -1,0 +1,28 @@
+defmodule Dialyxir.Output do
+  alias IO.ANSI
+
+  @warning_regex ~r/\w+.ex:\d+:/
+
+  def format(input) do
+    if ANSI.enabled? do
+      String.split(input, "\n")
+      |> Enum.map_join("\n", &colorize(&1))
+    else
+      input
+    end
+  end
+
+  defp colorize(line, color), do: color <> line <> ANSI.reset()
+
+  defp colorize("done (passed successfully)" = line), do: colorize(line, ANSI.green())
+
+  defp colorize("done (warnings were emitted)" = line), do: colorize(line, ANSI.yellow())
+
+  defp colorize(line) do
+    if String.match?(line, @warning_regex) do
+      colorize(line, ANSI.red())
+    else
+      line
+    end
+  end
+end
--- a/lib/mix/tasks/dialyzer.ex
+++ b/lib/mix/tasks/dialyzer.ex
@@ -138,19 +138,19 @@ defmodule Mix.Tasks.Dialyzer do
     not File.exists?(Project.deps_plt())
   end
 
-  defp dialyze(args, halt, ignore_warnings) do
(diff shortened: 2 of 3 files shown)


daveshah:
Thought I'd start the review and get your thoughts. If there's anything you think could use tweaking/change, or anything you think would be good to add, just let me know. Thanks! :)

lib/mix/tasks/dialyzer.ex
```diff
@@ -138,19 +138,19 @@ defmodule Mix.Tasks.Dialyzer do
...
   end
 
-  defp dialyze(args, halt, ignore_warnings) do
+  defp dialyze(args, halt, ignore_warnings, format \\ &Dialyxir.Output.format/1) do
     IO.puts "Starting Dialyzer"
     IO.puts "dialyzer " <> Enum.join(args, " ")
     {ret, exit_status} = System.cmd("dialyzer", args, [])
     exit_status = case ignore_warnings do
       nil ->
-        IO.puts ret
```
jeremyjh: We may want to restructure this existing code here so that we always print out each line at a time. I think presently you'd colorize all the output in the case that there are no ignore patterns. So only the filtering would be conditional, and we'd have just one place where we `IO.puts format.(line)` .

lib/dialyxir/output.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Dialyxir.Output do
+  @warning_regex ~r/\w+.ex:\d+:/
+
+  def format(input) do
+    String.split(input, "\n")
+    |> Enum.map_join("\n", &colorize(&1))
+  end
+
+  defp colorize(line) do
```
jeremyjh: We should check `IO.ANSI.enabled?` and just print without the escape codes if false.

lib/mix/tasks/dialyzer.ex
jeremyjh: Oh nevermind I missed you already handled this by splitting the lines in your formatter. So ignore this one :)

daveshah:
Made the update to check if ANSI was enabled. Good call! 👍 Also, while the PR's open, what are your thoughts on changing the color of the last line of the output so that 
`done (warnings were emitted)` shows up yellow and `done (passed successfully)` shows up green? Think that's overkill or should I add that in here too?

jeremyjh:
I think that would look nice actually; I'd say go ahead and add it if you'd like.

daveshah:
Awesome! Latest update has colors that look like the following:  

with warnings:  
screen shot 2017-01-17 at 9 45 38 am

With everything 👌 
screen shot 2017-01-17 at 9 46 17 am

PR #73, opened 2017-01-25
Use dialyzer API & add option for excluded_paths

### LINKED ISSUE #69: Using the :dialyzer.run API
Presently `dialyxir` shells out to the `dialyzer` command line program to perform analysis. This is undesirable because it imposes the cost of launching a whole VM just to perform analysis. This was not a conscious decision but simply a reflection of the fact that `dialyxir` began its life as the barest evolution of a shell script I'd been using. 

The [Erlang API](http://erlang.org/doc/man/dialyzer.html#gui-1) documentation is fairly straightforward. We have to build up a data structure containing all the required options and pass them to the `:dialyzer.run` function. 

There is a [working example](https://github.com/fishcakez/dialyze/blob/master/lib/mix/tasks/dialyze.ex#L435) in the dialyze project, the most relevant code:

```elixir
...
     plt_run([analysis_type: :succ_typings, plts: [plt], files: files,
          warnings: warnings])
   
  defp plt_run(opts) do
    :dialyzer.run([check_plt: false] ++ opts)
  end
```

## Warnings

You can see how they are structured at the bottom of the Erlang doc, you'll having a list of {:warning_name, {'filename', linenumber}, msg}.   `dialyzer` just prints them; of course we want to keep them colorized but it will be easier to have different color/styling for the the filename, line number, warning summary and the message itself.  

We can also group by module/file in some way visually, so that warnings from a single file visually fit together and down the line there might be something useful we could do knowing the categories of errors that user just saw; such as making it easier for them to find help info.

## Compatibility

1. By default the CLI returns warnings about unknown functions - keeping this the default means passing `:unknown` as one of the options.  
2. Presently users can pass flags to dialyzer either by including them a `flags: ["-Wno..."]` in their mix.exs or passing them on the command line. Any command line parameter that `dialyxir` isn't looking for just gets passed along to `dialyzer`. 

Number 2 is a little tricky but I think we have to map the flags to the atom names to do what the user means. e.g. we have to map "-Wno_unused" to `:no_unused` . We also want to let the user pass us atoms, e.g. they could have `dialyzer: flags: [:no_unused, :unmatched_returns]` in their Mix project.

If we have to break strict compatibility we can - we'd make this the primary feature of the `0.5` release. But we should do what we can to keep likely configurations working after the update.

daveshah:
Just as a heads up, this is still VERY MUCH work in progress. I finally set aside some time to work on this and thought I'd just push out changes as I go. If this gets too noisy, let me know. Also - if you have a WIP label I could apply to this, that might help denote it's WIP-ness :)

Thanks!

Closes #69


jeremyjh:
Ok, I will keep an eye on it but feel free to ping me for any feedback you'd like.

daveshah:
@jeremyjh - sorry this PR is taking so long! Been a busy last few weeks. I've been doing some refactoring on this one and if you think anything is out of place so far or have any concerns, feel free to mention them at any time :) I did noticed a recent change [here](https://github.com/jeremyjh/dialyxir/commit/a2851383bbcf2b6665260ff1d7c2f6544ab494bf) that I'm glad you added as we'll want to handle `:dialyzer.run` failures the same way in here.

daveshah:
@jeremyjh - I think this is in a good working spot and the items in the "Compatibility" section in #69 should be addressed. `-Woption` style warnings can still be passed in through either the command line (ex: `--optionname`, `--Woptionname`, etc...) or by including them as flags in mix.exs. Atom style flags should now work as well (`flags: ["-Woptionname", :optionname2, ...`)
All of these will just be passed in to the `warnings: [ ... ]` option in `dialyzer`. Let me know what you think and if there's anything you'd like changed.

Thanks! :dave

jeremyjh:
Hi Dave, I really appreciate the effort. Sorry I've been really busy with work work, so haven't had a chance to dig in yet but I do not see any immediate issues. It will probably be this weekend before I'm able to do a thorough review.

jeremyjh:
I was thinking we should pass :unknown as a default option, because in the command-line version by default unknown function warnings are displayed. But this begs the question, how would someone un-pass that option? In the command line version you cannot turn them off, this is probably a lot of what people use ignore-warnings for. I always thought there should be -Wno_unknown, but it would be silly to make that when we could just leave it as you have it, and document the fact that --unknown exists. Thoughts?

lib/mix/tasks/dialyzer.ex
```diff
@@ -130,6 +129,30 @@ defmodule Mix.Tasks.Dialyzer do
     end
   end
 
+  defp run_dialyzer(opts, dargs) do
+    raw_opts = Project.dialyzer_flags() ++ Enum.map(dargs, &elem(&1,0))
+    args = [ {:check_plt, false},
+             {:init_plt, String.to_charlist(Project.plt_file()) },
+             {:files_rec, Project.dialyzer_paths() },
+             {:warnings, transform(raw_opts) }]
```
daveshah: @jeremyjh , regarding your last comment:
We could pass this in by default to keep the same command-line version behavior and allow for a 'remove defaults' option (or something similar) in the config. Changing the above to something like:
```
{:warnings, transform(raw_opts) ++ @defaults -- user_removed() }]
```
Then we could document that users can just add `remove_defaults: [:unknown]` to their config if they'd like to change this behavior (instead of having to rely on `ignore-warnings` if this is what they're using it for)

I'd be cool w/making that change if you're up for it 👍

jeremyjh:
Yes that would work - I do think it is a better default to warn on unknown
and worth implementing.

lib/mix/tasks/dialyzer.ex
```diff
@@ -83,6 +83,7 @@ defmodule Mix.Tasks.Dialyzer do
   alias Dialyxir.Plt
   alias Dialyxir.Dialyzer
 
+  @default_warnings [ :unknown ]
```
daveshah: 👍 went ahead and added the default

lib/dialyxir/project.ex
```diff
@@ -42,6 +42,10 @@ defmodule Dialyxir.Project do
     paths -- excluded_paths |> Enum.map(&String.to_charlist/1)
   end
 
+  def dialyzer_removed_defaults do
+    dialyzer_config()[:remove_defaults] || []
+  end
```
daveshah: Also provided the means for removal 👍

mix.exs
```diff
@@ -11,7 +11,8 @@ defmodule Dialyxir.Mixfile do
       deps: [ {:ex_doc, ">= 0.0.0", only: :dev} ],
       dialyzer: [ plt_apps: [:dialyzer, :elixir, :kernel, :mix, :stdlib],
                   ignore_warnings: ".dialyzer_ignore",
-                  flags: ["-Wunmatched_returns", "-Werror_handling", "-Wunderspecs"]
+                  flags: ["-Wunmatched_returns", "-Werror_handling", "-Wunderspecs"],
+                  remove_defaults: [ :unknown ]
```
daveshah: @jeremyjh interestingly, by adding ` :unknown ` as a default to the warnings passed into `dialyzer.run`, we actually appeared to get different behavior in CI. I added the above line to make CI happy but just something to note as this may change the behavior for existing users.

jeremyjh:
Yes that is because the lines in .ignore-warnings no longer match the
output, though they mean the same thing. That is more of a general problem
people will have I think.

daveshah:
:man_facepalming: right! Cool 👍 Let me know if there's any other changes you think you'd like made :)

jeremyjh:
I've merged, going to let it bake a few days in master and see if we get any feedback before I push it to hex.

daveshah:
Sounds good 👍 I'll keep an eye out for feedback too and if any help is needed, I'll be happy to help where I can :)

jeremyjh:
@daveshah I went ahead and pushed this to `hex` as 0.5. Thanks SO much for all your effort on this and fixing the other bugs along the way, I really appreciate it.

daveshah:
`<high-five>` Anytime and definitely a ton of fun! :) `</high-five>`

PR #84, opened 2017-03-31
Make PLT bullet points render on github

vtsatskin:
Previously the bullet points in the PLT section looked like this on Github:

screen shot 2017-03-31 at 6 15 15 pm

Now they are formatted nicely.


jeremyjh:
Thanks!

PR #92, opened 2017-06-29
Remove deprecations for Elixir 1.5.0

odarriba:
Remove `()` for it's equivalent: `nil`.
This way it doesn't produce warnings by deprecation when using on Elixir 1.5.0


jeremyjh:
Thanks!

PR #94, opened 2017-07-17
Allow inclusion of blank lines in ignore file

mikestok:
I tried to use some blank lines to split up groups of lines in my file
of patterns to ignore, and discovered that if I added a blank line then
no filtering was done.

This is because `:binary.compile_pattern/1` treats an empty string in
the patterns as an argument error.

As a blank line isn't a meaningful thing to match using
String.contains?/2 (because all strings contain the empty string), it
seems to me that we can filter out blank lines in the ignore file so
that they can be used to separate chunks of patterns.


jeremyjh:
Thanks!

PR #95, opened 2017-07-25
chore(Elixir 1.5): Fix additional deprecation warnings for Elixir 1.5

tomciopp:
Fixes additional warnings when upgrading to Elixir 1.5


jeremyjh:
@tomciopp Hey, thanks for submitting this. 

One strange thing - Travis didn't build your PR - they flagged it as suspect. I checked with their support  -
 for this PR it does not really matter but it may affect other PR of yours in the future. They said it should be fixed if you login to https://travis-ci.org once. 

You could then push a commit to this branch to trigger a rebuild and confirm its working.

PR #99, opened 2017-08-18
Fix error on failed plt_build

tokafish:
When I was running mix dialyzer on a misconfigured project, I got the following error:

```** (Protocol.UndefinedError) protocol String.Chars not implemented for %IO.Stream{device: :standard_io, line_or_bytes: :line, raw: false}
    (elixir) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir) lib/string/chars.ex:22: String.Chars.to_string/1
    lib/mix/tasks/dialyzer.ex:179: Mix.Tasks.Dialyzer.build_parent_plt/0
    lib/mix/tasks/dialyzer.ex:113: Mix.Tasks.Dialyzer.no_check?/1
    lib/mix/tasks/dialyzer.ex:101: Mix.Tasks.Dialyzer.run/1
    (mix) lib/mix/task.ex:301: Mix.Task.run_task/3
    (mix) lib/mix/cli.ex:75: Mix.CLI.run_task/2
    (elixir) lib/code.ex:376: Code.require_file/2
```

I'm not sure what's changed with `IO.Stream` in recent versions of elixir, but this change fixes the problem (and the log of the failed dialyzer run still outputs)

### DIFF
--- a/lib/mix/tasks/dialyzer.ex
+++ b/lib/mix/tasks/dialyzer.ex
@@ -168,7 +168,7 @@ defmodule Mix.Tasks.Dialyzer do
 
   defp build_parent_plt() do
     parent = Mix.Project.config[:lockfile] |> Path.expand |> Path.dirname
-    opts = [ into: IO.stream(:stdio, :line),
+    opts = [ into: "",
              stderr_to_stdout: true,
              cd: parent ]
     # It would seem more natural to use Mix.in_project here to start in our parent project.


jeremyjh:
Can you tell me how to reproduce the problem?

When you tested successfully are you running dialyzer from the directory of an umbrella child?

tokafish:
Hey, thanks for your response!

Yeah, I was running from an umbrella child. In this case, the problem was that nerves uses subdirectories within deps in `mix.exs`, for example:

```
def project do
     [...
     deps_path: "../../deps/#{@target}",
     ...]
end
```

Where @target is set by an environment variable, which wasn't present when dialyzer shells out to generate the parent's plt. So the `mix dialyzer` that generates the parent's plt looks for deps in the wrong place, and errors. 
 
An easier way of reproducing is:

```
stream = IO.Stream(:stdio, :line)                                 
System.cmd("/bin/ls", [], [into: stream, stderr_to_stdout: true]) 
IO.puts("#{stream}")
```

So to summarize - the problem using dialyxir was with my setup, but this change should just get rid of the stack trace in the error message. Hope this helps!

jeremyjh:
Hi, I think what your simpler method tells us is that a Stream is not a String. But this does identify the problem:  I try to interpolate `out` into the error message. Since we redirected stderr to stdout and sent the cmd output to that stream `out` isn't going to contain anything we need at that point, and its obviously nothing we can use this way. The user should have already seen it. I think the proper fix is to ignore out and only supply `rc` in the response. I've got a branch almost ready to merge to support a quiet option and I've made this change [there](https://github.com/jeremyjh/dialyxir/pull/100/commits/efc18015b8d20431e64ce5c1027c7faa8aec5c7c).

jeremyjh:
Should be resolved in master.

PR #110, opened 2017-12-03
Update README.md for global install on Windows

ryanhart2:
Update section on installing globally with instruction specific to Windows as Windows has a different syntax for setting environment variables and executing multiple commands.


jeremyjh:
Thanks!

PR #100, opened 2017-08-20
Support a "Quiet" Option

jeremyjh:
This will resolve #85 .

Also resolves #111.

PR #111, opened 2017-12-31
Format code with the Elixir code formatter

ne1ro:
Hi!
First of all, thank you for the amazing library.

I want to suggest formatting `Dialyxir`'s code for upcoming Elixir 1.6 autoformatter. This will help us to keep the code style consistent across all files. The future changes should be also autoformatted.

**Changes**:

* `.formatter.exs` with initial configuration was added
* Code was formatted by `mix format` task

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,0 +1,9 @@
+export_locals_without_parens = []
+
+[
+  inputs: ["mix.exs", "lib/**/*.{ex,exs}", "test/**/*.{ex,exs}"],
+  locals_without_parens: export_locals_without_parens,
+  export: [
+    locals_without_parens: export_locals_without_parens
+  ]
+]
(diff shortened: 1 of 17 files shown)


jeremyjh:
Hi, I think this is a good thing to do and thanks for submitting this. Since it touches so many lines though, I'm going to have to solve 30+ merge conflicts when I finish up and merge #100. That's been stalled for a bit but will be done. I'd rather just re-run the tool after that is done, or if you like you could resubmit this against the `quiet` branch. If you do that please add the locals `error` and `info` to `locals_without_parens` as I prefer those the way they are.

ne1ro:
Hello, @jeremyjh , sorry for the late response.

I've added `error/1` and `info/1` to `locals_without_parens`, but I suppose we need to wait for https://github.com/jeremyjh/dialyxir/pull/100 to be in the master branch and then I will re-run the formatter.

Issue #107, opened 2017-11-28
Add dialyzer.clean task


christhekeele:
Adding the ability to `rm` all plts by running `mix dialyzer.clean` would be very useful, especially in CI environments.

This also keeps end users from having to know where `dialyxir` happens to place its plts for any given version (even though this is not likely to change).

Finally, if other `mix dialyzer` task-adding tools do the same, it creates a really nice API where you don't even have to know which tool you are using to purge plts. To that end I've duplicated this feature request here as well: Comcast/dialyzex#1.

jeremyjh:
I agree its a good idea.

jeremyjh:
@christhekeele Do you picture any arguments being relevant to this task? For example should it be possible to clean just both the current configuration (current elixir, erlang version PLTs) versus PLTs from all versions (e.g. `--all`). I think if there are not any command line switches specific to this task, then this task itself should just be a command-line switch (similar to `--plt`). This is to avoid needless pollution of the `mix help` space.

christhekeele:
- I can't think of any relevant generic flags, and specific ones (ie what gets cleaned) are inherently dialyxir-implementation specific
- I feel as if the flag-based approach is less generic than the task-based one, which defeats some of the cross-dialyzer-task-parity I had in mind for this
- The task needn't appear in `mix help` if it lacks a description, so long as it is well documented within the project

YMMV, though, but that's what I've been envisioning: a simple, single, generic cross-tool API.

jeremyjh:
I don't think a cross-tool API will really be of much benefit to anyone right now, and isn't really worth pursuing as a design goal. Where it doesn't have any other costs I'm not opposed to it though.

Its an interesting idea to hide the task via doc, but I wonder if this would make the documentation a bit awkward as we'd really need to document `dialyzer.clean` in the `dialyzer` task's doc.  Maybe burying the task is a problem too though; the `--plt` flag seems to escape a lot of people's notice.

Right now I'm leaning towards implementing this under a flag `--clean` to be consistent with `--plt`.

christhekeele:
I do feel as if `--clean` is more consistent with the `--plt` flag today, although I'll admit I'm not a huge fan of using flags in lieu of subtasks. In my opinion subtasks are easier to invoke programmatically and flags should be for configuration rather than behaviour selection. This is completely personal preference, though, I'm happy to have just the flags should that be your preference.

jeremyjh:
[...]

PR #116, opened 2018-03-21
Added dialyxir.clean task

gnufede:
Added `mix dialyxir.clean` task that deletes PLT and hash files.

Added a new parameter to `find_plts` function, with the function to be called
on the PLTs found.

Solves [https://github.com/jeremyjh/dialyxir/issues/107](https://github.com/jeremyjh/dialyxir/issues/107)

### DIFF
--- a/lib/dialyxir/plt.ex
+++ b/lib/dialyxir/plt.ex
@@ -9,27 +9,32 @@ defmodule Dialyxir.Plt do
 
   def check(plts) do
     info("Finding suitable PLTs")
-    find_plts(plts, [])
+    find_plts(plts, [], &check_plt/4)
   end
 
-  defp find_plts([{plt, apps} | plts], acc) do
+  def clean(plts) do
+    info("Deleting PLTs")
+    find_plts(plts, [], &delete_plt/4)
+  end
+
+  defp find_plts([{plt, apps} | plts], acc, fun) do
     case plt_files(plt) do
       nil ->
-        find_plts(plts, [{plt, apps, nil} | acc])
+        find_plts(plts, [{plt, apps, nil} | acc], fun)
       beams ->
         apps_rest = Enum.flat_map(plts, fn({_plt2, apps2}) -> apps2 end)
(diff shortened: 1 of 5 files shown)


jeremyjh:
I think this looks good but we also need to add docs so that something is displayed for the task in `mix help` as well as `mix help dialyzer.clean`.  Also this is probably pretty testable - could you look at also adding tests for this as well please?

jeremyjh:
Hi, sorry I neglected reviewing this further until now. I took a look and everything you have looks fine, but re-reading #107 again I'm left with the impression it was also requesting the core plts (in MIX_HOME) would be deleted. I'm not sure why that would be necessary though -  @christhekeele can you clarify if that was the original ask, and if so what benefit it would provide?

christhekeele:
That was the original ask—though I think the partial purge might be useful to someone, perhaps as a flag.

What I was requesting is the ability to fully unwind any artifacts produced by `mix dialyzer` tasks, without having to know where it stores them. This is to serve the same purpose as `mix deps.clean` and `mix clean`: to blow away all state so you know that `mix xxx.build` starts with a fresh state.

The reasoning is that while common usecase won't need this, if you're wrangling CI environments or trying to build PLTs in advance, you're probably doing something bespoke enough that you want full control over all files the task touches.

jeremyjh:
@christhekeele the difference to `mix deps.clean` or `mix clean` is that their scope is limited to the current project directory. Deleting shared files in MIX_HOME might not be the first thing you'd expect from this task, it would seem like that would be safer to put behind an `--all` flag.

christhekeele:
I agree that an `--all` flag makes sense for this, and has nice parity with similar tasks (`mix deps.clean --all` for instance).

gnufede:
Agreed, I'll make the changes so it'll only delete core plts if `--all` flag is set

gnufede:
I think it's ready now @jeremyjh, I hope you like the approach.

jeremyjh:
Looks great, thanks @gnufede !

PR #118, opened 2018-04-24
WIP - Elixir formatter for Dialyzer warnings

### LINKED ISSUE #103: Provide better error messages
Hi folks, first off, thanks for Dialyxir! I would like to request a feature. :)

Dialyzer error messages are not good. They are very cryptic and hard to understand. When integrating via Elixir is even worse, as terms are formatted in Erlang syntax. Good news is: it is not our fault and we can do better!

We already receive all of the warnings from dialyzer [here](https://github.com/jeremyjh/dialyxir/blob/2a72cba00daf56474287ab3a5b95fcefb870b149/lib/dialyxir/dialyzer.ex#L23). So we can take the liberty of formatting them in better ways. [Rebar3 does that](https://github.com/erlang/rebar3/blob/master/src/rebar_dialyzer_format.erl). Just don't forget the catch all clause with a behaviour that is similar to today's.

How should the error messages look like? This [Elm article](http://elm-lang.org/blog/compiler-errors-for-humans) is an excellent guideline but we need to say what went wrong, why it went wrong and propose next steps.

The Credo folks are also good on their error messages so it can also be a good starting point.

FWIW, Elixir does a very similar thing to what I am proposing here, except we are handling the warnings coming from compiler.

Although I cannot directly work on this feature, I will be glad to provide guidance. Folks have mentioned the difficulty in handling those error messages in the past so I thought I would pass the massage forward.

asummers:
Resolves #103 -- See discussion there for history/context. @jeremyjh suggested I open this to get the cycle started, so here we go!

NOTE: This is still WIP. There is a bunch left to do, but this should hopefully get the ball rolling and we can collect as much feedback from real life use cases as possible.

This PR adds a basic lexer and parser using the built in Erlang tools that allows translation from Erlang formatted errors to Elixir. It does some dirty tricks to get the formatter to make things look pretty, and I'm a little ashamed of what I did there. Any suggestions on how to improve that would be welcome. 

TODO:
- [ ] Add "Short Error" for errors [See 1]
- [ ] Add filtering logic for short errors
- [ ] Clean up shaky lexer/parser rules [See 2]
- [ ] Test on as much real life input as possible
- [ ] Improve error messages wherever possible
- [ ] Add fallback if the pretty printing/parsing fails.
- [ ] Documentation

-- 

1) Short Error - We need to have a way to export the newly verbose error message to things like `.dialyzer.ignore_warnings` files and similar, as well as providing short messages for use in e.g. text editors. 

2) The rules were done on input I had access to, and I don't tend to keep a ton of Dialyzer error laden projects at hand. This means that the rules are ignoring all sorts of input. For example, any non a-Z characters in atoms will currently fail. I don't have much access to internationalized code, so I can't test this very easily. Any samples of failing input  are greatly appreciated. There's probably even much more basic constructs that I'm missing. 

These are the things still left to do off the top of my head, but I may be missing others.


jeremyjh:
I think we should crowd-source the problem of building a corpus of test data. I'm thinking of making aforum post requesting text output samples - but we'll have to reverse engineer the term format warnings from those. Is there any better alternative to that? Mostly we just need the type examples, which is text anyway.  It is too bad I didn't implement `--raw` yet.

asummers:
The best way I've found to debug is that when it chokes on parsing I want the raw value... when I then REPL debug til it parses so maybe I just build that in? Like try to parse and then fail gracefully with bug report instructions? I noticed it chokes on structs of the form ``#{'__struct__':='module', 'arg1':='arg2', _ => _}` last night (because why NOT change the arrows!), so I'm going to work on adding that and I'll dovetail that into a nice error message with the raw output.  Then we can get actionable bug reports and build up the test suite at the same time. Crowd sourcing this is a good idea :)

asummers:
I'm thinking something like this:

```
Failed to parse part of warning:
{:warn_matching, {'apps/mandark/lib/web/endpoint.ex', 1}, {:pattern_match, ['pattern _@22 = \#{\'halted\':=\'true\', \'__struct__\':=\'Elixir.Plug.Conn\'}', '\'error\' | \'ok\'']}}

Failing part:
pattern _@22 = #{'halted':='true', '__struct__':='Elixir.Plug.Conn'}

Please file a bug with this message.

Legacy warning:
apps/mandark/lib/web/endpoint.ex:1: The pattern _@22 = #{'halted':='true', '__struct__':='Elixir.Plug.Conn'} can never match the type 'error' | 'ok'
```

asummers:
Some more complicated example output:

```
Expression produces a value of type %Plug.Conn{
  :adapter => {atom(), _},
  :assigns => %{atom() => _},
  :before_send => [
    (%Plug.Conn{
       :adapter => {_, _},
       :assigns => map(),
       :before_send => [(_ -> any())],
       _ => _
     } ->
       %Plug.Conn{
         :adapter => {_, _},
         :assigns => map(),
         :before_send => [(_ -> any())],
         _ => _
       })
  ],
  :body_params => %Plug.Conn.Unfetched{
    :aspect => atom(),
    binary() =>
      binary()
      | [binary() | [any()] | %{binary() => _}]
      | %{binary() => binary() | [any()] | %{binary() => _}}
  },
  :cookies => %Plug.Conn.Unfetched{:aspect => atom(), binary() => binary()},
  :halted => true,
  :host => binary(),
  :method => binary(),
  :owner => pid(),
  :params => %Plug.Conn.Unfetched{
    :aspect => atom(),
    binary() =>
      binary()
      | [binary() | [any()] | %{binary() => _}]
      | %{binary() => binary() | [any()] | %{binary() => _}}
  },
  :path_info => [binary()],
  :path_params => %{
    binary() =>
      binary()
      | [binary() | [any()] | %{binary() => _}]
      | %{binary() => binary() | [any()] | %{binary() => _}}
  },
  :peer =>
    {{byte(), byte(), byte(), byte()}
     | {char(), char(), char(), char(), char(), char(), char(), char()}, char()},
  :port => char(),
  :private => %{atom() => _},
  :query_params => %Plug.Conn.Unfetched{
    :aspect => atom(),
    binary() =>
      binary()
      | [binary() | [any()] | %{binary() => _}]
      | %{binary() => binary() | [any()] | %{binary() => _}}
  },
  :query_string => binary(),
  :remote_ip =>
    {byte(), byte(), byte(), byte()}
    | {char(), char(), char(), char(), char(), char(), char(), char()},
  :req_cookies => %Plug.Conn.Unfetched{:aspect => atom(), binary() => binary()},
  :req_headers => [{binary(), binary()}],
  :request_path => binary(),
  :resp_body =>
    nil
    | binary()
    | maybe_improper_list(
        binary() | maybe_improper_list(any(), binary() | []) | byte(),
        binary() | []
      ),
  :resp_cookies => %{binary() => %{}},
  :resp_headers => [{binary(), binary()}],
  :scheme => :http | :https,
  :script_name => [binary()],
  :secret_key_base => nil | binary(),
  :state => :chunked | :file | :sent | :set | :set_chunked | :set_file | :unset,
  :status => nil | non_neg_integer()
}, but this value is unmatched.
________________________________________________________________________________
```

asummers:
@jeremyjh I added behavior for graceful fallback of both missing warnings (which I found a few of) and parse failures. Should I direct them here or have them make discrete bugs? Probably worth making a forum post at this point.

jeremyjh:
This is great. I pushed a copy of your branch to this repo. I'll write something up and post it in the morning and include instructions for using this branch. I think we'll want people to post any reports in this PR for now.

jeremyjh:
I found one parsing issue but it doesn't throw an exception. I think a `--raw` option which dumps the term literals would be useful in cases like this, and maybe others. It could also be helpful to generate messages to use in test cases. So I implemented that in my clone of this branch.

I think we will want to fix this issue though before we request feedback, it seems a naked map is not getting parsed. 

The case I found is the `pattern` part of  a `:pattern_match` warning for example I get: 

```
The :pattern can never match the type %Diatarget.AStruct{:field_one => <<_::40>>, :field_two => 0}.
```

From the raw message, the text that cannot be parsed is  ` #{\'field_three\':=_@1}`

```elixir
{:warn_matching, {'lib/diatarget.ex', 32}, {:pattern_match, ['pattern \#{\'field_three\':=_@1}', '\#{\'__struct__\':=\'Elixir.Diatarget.AStruct\', \'field_one\':=<<_:40>>, \'field_two\':=0}']}}
```
I also tried `pretty_print_type` but I'm thinking the formatter must not handle naked maps yet.

Sample code: 

```elixir
  defmodule AStruct do
    defstruct field_one: "hello", field_two: 0
  end

  def bad_access do
    %AStruct{}.field_three
  end
```

asummers:
@jeremyjh That's the `pattern` part that's failing. It needs to drop the `pattern ` part from the string, I think.

zachdaniel:
Not sure how possible this would be, but I think the next big win with formatting the output would be minimization to only show the differences. If one member of a map is typed differently then expected, it would be really great if we could reduce the noise during formatting. Instead of seeing two giant maps we could see a message like `member :id of map does not meet the specification`

asummers:
@zachdaniel There's a little bit of discussion about that in the issue. The short answer is "eventually" but the long answer is that in order to do that, we have to semantically parse this input, and then apply some intelligent reasoning at the level where we're currently just calling `pretty_print*/1` in `formatter.ex`.  Dialyzer has different levels of detail that it gives you, in the same warning, so we'd have to develop rules about when it's safe to do that kind of thing. I wish we could just apply diffs and call it a day 😦

amatalai:
Hi, thanks for great work.

I started to test it in the project with a large use of dialyzer and so far, I have encountered this*:

```
Please file a bug with this message.

Failed to parse part of warning:
{:warn_failing_call, {'apps/my_app/lib/my_app/fetch/thread.ex', 3}, {:call, [MyApp.Fetch.Thread, :pipeline, '(_@3::\#{\'__struct__\':=\'Elixir.MyApp.Fetch.Thread.Data\', \'admin\':=_, \'thread\':=\'nil\', \'thread_id\':=integer()})', [1], :both, '(\#{\'__struct__\':=\'Elixir.MyApp.Fetch.Thread.Data\', \'admin\':=\#{\'__meta__\':=_, \'__struct__\':=\'Elixir.Storage.Admin\', \'active_thread_number\':=_, \'x\':=_, \'avatar_url\':=_, \'deleted\':=_, \'fake\':=_, \'first_name\':=_, \'external_id\':=_, \'id\':=_, \'inserted_at\':=_, \'last_name\':=_, \'supervisor\':=_, \'threads\':=_, \'title\':=_, \'updated_at\':=_}, \'thread\':=\#{\'__meta__\':=_, \'__struct__\':=\'Elixir.Storage.Thread\', \'admin\':=_, \'admin_id\':=_, \'things\':=_, \'completed\':=_, \'failed_ids\':=_, \'metadata\':=_, \'id\':=_, \'inserted_at\':=_, \'last_message_at\':=_, \'messages\':=_, \'not_answered\':=_, \'list_field_b\':=_, \'list_field_a\':=_, \'purpose\':=_, \'ready\':=_, \'unread_count\':=_, \'unresolved\':=_, \'updated_at\':=_, \'user\':=_, \'user_id\':=_}, \'thread_id\':=integer()})', '{\'error\',\'not_found\'} | {\'ok\',\#{\'__meta__\':=_, \'__struct__\':=\'Elixir.Storage.Thread\', \'admin\':=_, \'admin_id\':=_, \'things\':=_, \'completed\':=_, \'failed_ids\':=_, \'metadata\':=_, \'id\':=_, \'inserted_at\':=_, \'last_message_at\':=_, \'messages\':=_, \'not_answered\':=_, \'list_field_b\':=_, \'list_field_a\':=_, \'purpose\':=_, \'ready\':=_, \'unread_count\':=_, \'unresolved\':=_, \'updated_at\':=_, \'user\':=_, \'user_id\':=_}}', {true, '(data::\'Elixir.MyApp.Fetch.Thread.Data\':t()) -> {\'ok\',\#{\'__struct__\':=\'Elixir.Storage.Thread\',\'__meta__\':=term(),\'admin\':=term(),\'admin_id\':=term(),\'things\':=term(),\'completed\':=term(),\'failed_ids\':=term(),\'metadata\':=term(),\'id\':=term(),\'inserted_at\':=term(),\'last_message_at\':=term(),\'messages\':=term(),\'not_answered\':=term(),\'list_field_b\':=term(),\'list_field_a\':=term(),\'purpose\':=term(),\'ready\':=term(),\'unread_count\':=term(),\'unresolved\':=term(),\'updated_at\':=term(),\'user\':=term(),\'user_id\':=term()}} | {\'error\',\'not_found\'}'}]}}

Failing part:
(data::'Elixir.MyApp.Fetch.Thread.Data':t()) -> {'ok',#{'__struct__':='Elixir.Storage.Thread','__meta__':=term(),'admin':=term(),'admin_id':=term(),'things':=term(),'completed':=term(),'failed_ids':=term(),'metadata':=term(),'id':=term(),'inserted_at':=term(),'last_message_at':=term(),'messages':=term(),'not_answered':=term(),'list_field_b':=term(),'list_field_a':=term(),'purpose':=term(),'ready':=term(),'unread_count':=term(),'unresolved':=term(),'updated_at':=term(),'user':=term(),'user_id':=term()}} | {'error','not_found'}

Legacy warning:
apps/my_app/lib/my_app/fetch/thread.ex:3: The call 'Elixir.MyApp.Fetch.Thread':pipeline(_@3::#{'__struct__':='Elixir.MyApp.Fetch.Thread.Data', 'admin':=_, 'thread':='nil', 'thread_id':=integer()}) will never return since the success typing is (#{'__struct__':='Elixir.MyApp.Fetch.Thread.Data', 'admin':=#{'__meta__':=_, '__struct__':='Elixir.Storage.Admin', 'active_thread_number':=_, 'x':=_, 'avatar_url':=_, 'deleted':=_, 'fake':=_, 'first_name':=_, 'external_id':=_, 'id':=_, 'inserted_at':=_, 'last_name':=_, 'supervisor':=_, 'threads':=_, 'title':=_, 'updated_at':=_}, 'thread':=#{'__meta__':=_, '__struct__':='Elixir.Storage.Thread', 'admin':=_, 'admin_id':=_, 'things':=_, 'completed':=_, 'failed_ids':=_, 'metadata':=_, 'id':=_, 'inserted_at':=_, 'last_message_at':=_, 'messages':=_, 'not_answered':=_, 'list_field_b':=_, 'list_field_a':=_, 'purpose':=_, 'ready':=_, 'unread_count':=_, 'unresolved':=_, 'updated_at':=_, 'user':=_, 'user_id':=_}, 'thread_id':=integer()}) -> {'error','not_found'} | {'ok',#{'__meta__':=_, '__struct__':='Elixir.Storage.Thread', 'admin':=_, 'admin_id':=_, 'things':=_, 'completed':=_, 'failed_ids':=_, 'metadata':=_, 'id':=_, 'inserted_at':=_, 'last_message_at':=_, 'messages':=_, 'not_answered':=_, 'list_field_b':=_, 'list_field_a':=_, 'purpose':=_, 'ready':=_, 'unread_count':=_, 'unresolved':=_, 'updated_at':=_, 'user':=_, 'user_id':=_}} and the contract is (data::'Elixir.MyApp.Fetch.Thread.Data':t()) -> {'ok',#{'__struct__':='Elixir.Storage.Thread','__meta__':=term(),'admin':=term(),'admin_id':=term(),'things':=term(),'completed':=term(),'failed_ids':=term(),'metadata':=term(),'id':=term(),'inserted_at':=term(),'last_message_at':=term(),'messages':=term(),'not_answered':=term(),'list_field_b':=term(),'list_field_a':=term(),'purpose':=term(),'ready':=term(),'unread_count':=term(),'unresolved':=term(),'updated_at':=term(),'user':=term(),'user_id':=term()}} | {'error','not_found'}
```

\* I changed some names that I should not share on the internet

asummers:
Interesting. So it sounds like it's failing on errors when you say something along the lines of 

```
@spec frobincate(foo :: Foo.t(), bar :: Bar.t()) :: Baz.t()
```

@amatalai Can you confirm that suspicion in your source?

amatalai:
Yes, it happens only for named arg in typespec. The problem disappeared after changing
```elixir
@spec function_name(data :: Data.t()) ::
          {:ok, %Storage.Thread{}}
          | {:error, :not_found}
```
to
```elixir
@spec function_name(Data.t()) ::
          {:ok, %Storage.Thread{}}
          | {:error, :not_found}
```

asummers:
Much appreciated 😃

asummers:
@amatalai Fixed. Once we figure out the best workflow for getting that into the branch here, it'll be available. Thank you for the report 😄 ♥️

amatalai:
Found one more

Sample code:
```elixir
defmodule FunctionArg do
  @spec a(integer, (() -> integer)) :: integer
  def a(x, fun) do
    x + fun.()
  end

  def b do
    a(:ok, fn -> 5 end)
  end
end
```
mix dialyzer output:
```
lib/function_arg.ex:7
Function b/0 has no local return.
________________________________________________________________________________
Please file a bug with this message.

Failed to parse part of warning:
{:warn_failing_call, {'lib/function_arg.ex', 8}, {:call, [FunctionArg, :a, '(\'ok\',fun(() -> 5))', [1], :both, '(number(),fun(() -> any()))', 'number()', {true, '(integer(),fun(() -> integer())) -> integer()'}]}}

Failing part:
('ok',fun(() -> 5))

Legacy warning:
lib/function_arg.ex:8: The call 'Elixir.FunctionArg':a('ok',fun(() -> 5)) will never return since the success typing is (number(),fun(() -> any())) -> number() and the contract is (integer(),fun(() -> integer())) -> integer()
________________________________________________________________________________
```

The problem occurs only when function passed as arg is zero arity

asummers:
@amatalai  Fixed when the above PR gets merged. Thank you 😄 ♥️ 💚 

```
(:ok, (() -> 5))
```

amatalai:
Issue with strings when function takes more than 1 arg

Sample code: 
```elixir
defmodule StringIssue do
  def single_arg do
    do_single_arg("S")
  end

  def multiple_args do
    do_multiple_args("S", 12345)
  end

  defp do_single_arg(var1) when is_atom(var1), do: :ok
  defp do_multiple_args(var1, _var2) when is_atom(var1), do: :ok
end
```
mix dialyzer output:
```
lib/string_issue.ex:2
Function single_arg/0 has no local return.
________________________________________________________________________________
lib/string_issue.ex:3
The call StringIssue.do_single_arg(<<_::8>>) will never return since it differs in argument positions 1 from the success typing arguments: (atom()).
________________________________________________________________________________
lib/string_issue.ex:6
Function multiple_args/0 has no local return.
________________________________________________________________________________
Please file a bug with this message.

Failed to parse part of warning:
{:warn_failing_call, {'lib/string_issue.ex', 7}, {:call, [StringIssue, :do_multiple_args, '(\#{#<83>(8, 1, \'integer\', [\'unsigned\', \'big\'])}#,12345)', [1], :only_sig, '(atom(),any())', '\'ok\'', {false, :none}]}}

Failing part:
(#{#<83>(8, 1, 'integer', ['unsigned', 'big'])}#,12345)

Legacy warning:
lib/string_issue.ex:7: The call 'Elixir.StringIssue':do_multiple_args(#{#<83>(8, 1, 'integer', ['unsigned', 'big'])}#,12345) will never return since it differs in the 1st argument from the success typing arguments: (atom(),any())
________________________________________________________________________________
lib/string_issue.ex:10
Function do_single_arg/1 has no local return.
________________________________________________________________________________
lib/string_issue.ex:10
Guard test is_atom(Vvar1@1::<<_:8>>) can never succeed.
________________________________________________________________________________
lib/string_issue.ex:11
Function do_multiple_args/2 has no local return.
________________________________________________________________________________
lib/string_issue.ex:11
Guard test is_atom(Vvar1@1::<<_:8>>) can never succeed.
________________________________________________________________________________
```

`#{#<83>(8, 1, 'integer', ['unsigned', 'big'])}#` is kind of wtf for me

Additional info:
* erlang 20.2, elixir 1.6.4-otp-20 (both installed with asdf version manager)
* dialyzer config: `[flags: [:error_handling, :race_conditions, :underspecs, :unmatched_returns]]`

asummers:
Hmm. That's a strange one. I'll look into it and figure it out ☺️

Out of curiosity how is the output otherwise, since you've been working with it? Anything you wish was different?

asummers:
Report from https://github.com/jeremyjh/dialyxir/issues/122

```
________________________________________________________________________________
Please file a bug with this message.

Failed to parse part of warning:
{:warn_matching, {'lib/eaten_web/controllers/session_controller.ex', 6}, {:pattern_match, ['pattern {\'ok\', Vuser@1}', '{\'error\',<<_:64,_:_*8>>}']}}

Failing part:
{'error',<<_:64,_:_*8>>}

Legacy warning:
lib/eaten_web/controllers/session_controller.ex:6: The pattern {'ok', Vuser@1} can never match the type {'error',<<_:64,_:_*8>>}
________________________________________________________________________________
```

NobbZ:
In the other issue, I was asked to get into detail about the specs. The function called is not speced, nor is any other in the callchain.

I assume it is inferred. Basically I am calling Comeonin.Argon2.check_pw.

I will try to create a more minimal example when I am back home. Can take several hours though.

asummers:
@NobbZ that's enough for me to work with, I think. I'll look into it this evening. Thank you!

zachdaniel:
Running into an issue with this branch: 

```
** (FunctionClauseError) no function clause matching in :dialyzer.format_warning/2

    The following arguments were given to :dialyzer.format_warning/2:

        # 1
        {:unknown_type, {Ecto.Query, :dynamic_expr, 0}}

        # 2
        :fullpath
```

asummers:
@zachdaniel Should be fixed now. Unsure how Dialyzer itself was complaining there about its own error, but the branch here now supports all the warnings from Dialyzer code. I was missing a few.

nikolauska:
When dialyzer fails on Plug.Conn check this error will be thrown
```
"{atom(), _}"
"{_, _}"
"{_, _}"
"{byte(), byte(), byte(), byte()}"
"{char(), char(), char(), char(), char(), char(), char(), char()}"
"{{byte(), byte(), byte(), byte()} | {char(), char(), char(), char(), char(), char(), char(), char()}, char()}"
"{byte(), byte(), byte(), byte()}"
"{char(), char(), char(), char(), char(), char(), char(), char()}"
"{binary(), binary()}"
"{binary(), binary()}"
** (SyntaxError) nofile:1: unexpected token: ":" (column 1965, codepoint U+003A)
    (elixir) lib/code.ex:442: Code.format_string!/2
    lib/dialyxir/pretty_print.ex:64: Dialyxir.PrettyPrint.pretty_print_args/1
    lib/dialyxir/formatter.ex:92: Dialyxir.Formatter.message_to_string/1
    lib/dialyxir/formatter.ex:66: Dialyxir.Formatter.format_warning/2
    lib/dialyxir/formatter.ex:21: anonymous fn/3 in Dialyxir.Formatter.format_and_filter/3
    (elixir) lib/enum.ex:1294: Enum."-map/2-lists^map/1-0-"/2
    (elixir) lib/enum.ex:1294: Enum."-map/2-lists^map/1-0-"/2
    lib/dialyxir/formatter.ex:18: Dialyxir.Formatter.format_and_filter/3
    lib/dialyxir/dialyzer.ex:15: Dialyxir.Dialyzer.Runner.run/2
    lib/dialyxir/dialyzer.ex:31: Dialyxir.Dialyzer.dialyze/3
    lib/mix/tasks/dialyzer.ex:203: Mix.Tasks.Dialyzer.run_dialyzer/2
    (mix) lib/mix/task.ex:314: Mix.Task.run_task/3
    (mix) lib/mix/cli.ex:80: Mix.CLI.run_task/2
    (elixir) lib/code.ex:677: Code.require_file/
```

Code it is trying to format looks like this:
```
@spec a(%Plug.Conn{:adapter => {atom(), _}, :assigns => %{atom() => _}, :before_send => [((%Plug.Conn{:adapter => {_, _}, :assigns => map(), :before_send => [((_) -> any())], _ => _}) -> %Plug.Conn{:adapter => {_, _}, :assigns => map(), :before_send => [((_) -> any())], _ => _})], :body_params => %Plug.Conn.Unfetched{:aspect => atom(), binary() => binary() | [binary() | [any()] | %{binary() => _}] | %{binary() => binary() | [any()] | %{binary() => _}}}, :cookies => %Plug.Conn.Unfetched{:aspect => atom(), binary() => binary()}, :halted => _, :host => binary(), :method => binary(), :owner => pid(), :params => %Plug.Conn.Unfetched{:aspect => atom(), binary() => binary() | [binary() | [any()] | %{binary() => _}] | %{binary() => binary() | [any()] | %{binary() => _}}}, :path_info => [binary()], :path_params => %{binary() => binary() | [binary() | [any()] | %{binary() => _}] | %{binary() => binary() | [any()] | %{binary() => _}}}, :peer => {{byte(), byte(), byte(), byte()} | {char(), char(), char(), char(), char(), char(), char(), char()}, char()}, :port => char(), :private => %{atom() => _}, :query_params => %Plug.Conn.Unfetched{:aspect => atom(), binary() => binary() | [binary() | [any()] | %{binary() => _}] | %{binary() => binary() | [any()] | %{binary() => _}}}, :query_string => binary(), :remote_ip => {byte(), byte(), byte(), byte()} | {char(), char(), char(), char(), char(), char(), char(), char()}, :req_cookies => %Plug.Conn.Unfetched{:aspect => atom(), binary() => binary()}, :req_headers => [{binary(), binary()}], :request_path => binary(), :resp_body => nil | binary() | maybe_improper_list(binary() | maybe_improper_list(any(), binary() | []) | byte(), binary() | []), :resp_cookies => %{binary() => %{}}, :resp_headers => [{binary(), binary()}], :scheme => :http | :https, :script_name => [binary()], :secret_key_base => nil | binary(), :state => :chunked | :file | :sent | :set | :set_chunked | :set_file | :unset, :status => nil | :1..1114111}, MyApp.ErrorView, <<_ :: 112>>) :: :ok
def a() do
  :ok
end
```

asummers:
@nikolauska Thank's for doing the legwork on that. Ranges seem to be getting printed with a leading `:` for some reason.

amatalai:
Dialyxir branch: `elixir-formatter`, ref: `b0c63083f69c90caeadb2545596d9fc9c7820beb`

Sample code:
```
defmodule RepeatedVarNameInWith do
  def x do
    data = 1

    with {:ok, data} <- a(data),
         {:ok, data} <- b(data),
         {:ok, _data} <- c(data) do
      :ok
    end
  end

  defp a(int), do: {:ok, int}
  defp b(int) when is_binary(int), do: {:ok, int}
  defp c(int), do: {:ok, int}
end
```

mix dialyzer output:
```
** (SyntaxError) nofile:1: invalid character "@" (codepoint U+0040) in alias: Vdata@2
    (elixir) lib/code.ex:442: Code.format_string!/2
    lib/dialyxir/pretty_print.ex:64: Dialyxir.PrettyPrint.pretty_print_args/1
    lib/dialyxir/formatter.ex:92: Dialyxir.Formatter.message_to_string/1
    lib/dialyxir/formatter.ex:66: Dialyxir.Formatter.format_warning/2
    lib/dialyxir/formatter.ex:21: anonymous fn/3 in Dialyxir.Formatter.format_and_filter/3
    (elixir) lib/enum.ex:1294: Enum."-map/2-lists^map/1-0-"/2
    (elixir) lib/enum.ex:1294: Enum."-map/2-lists^map/1-0-"/2
```

mix dialyzer --raw output:
```
{:warn_return_no_exit, {'lib/same_var_name_in_with.ex', 2}, {:no_return, [:only_normal, :x, 0]}}
{:warn_failing_call, {'lib/same_var_name_in_with.ex', 6}, {:call, [SameVarNameInWith, :b, '(Vdata@2::1)', [1], :only_sig, '(binary())', '{\'ok\',binary()}', {false, :none}]}}
{:warn_return_no_exit, {'lib/same_var_name_in_with.ex', 13}, {:no_return, [:only_normal, :b, 1]}}
{:warn_matching, {'lib/same_var_name_in_with.ex', 13}, {:guard_fail, [:is_binary, '(Vint@1::1)']}}
{:warn_not_called, {'lib/same_var_name_in_with.ex', 14}, {:unused_fun, [:c, 1]}}
```

Issue is caused by `'(Vdata@2::1)'` in 2nd line or `'(Vint@1::1)'` in 4th line

amatalai:
>Thank you. It seems that's how Elixir gets around Erlang's single variable assignment rules, by just making a new variable V#{shadowed_named}::#{shadow_number} and I'm assuming V for variable. Clever. Interesting bug 😄

Yes, there's a lot of those when you decompile .beam file to erlang

Btw string issue still occurs for strings with length != 1. #127 should help

amatalai:
Missing unicode support in atoms
branch: `elixir-formatter` ref: `fdb169e5b12a7dfc80e89ea577bd97f9795de3ee`

Sample code:
```elixir
defmodule UnicodeSupportInAtoms do
  def x do
    :other = :".$@\data\u{23F0}!"
  end
end
```

mix dialyzer output:
```
lib/unicode_support_in_atoms.ex:2
Function x/0 has no local return.
________________________________________________________________________________
Please file a bug in https://github.com/jeremyjh/dialyxir/pull/118 with this message.

Failed to parse part of warning:
{:warn_matching, {'lib/unicode_support_in_atoms.ex', 3}, {:pattern_match, ['pattern \'other\'', [39, 46, 36, 64, 92, 100, 97, 116, 97, 9200, 33, 39]]}}

Failing part:
'.$@\data⏰!'

Legacy warning:
lib/unicode_support_in_atoms.ex:3: The pattern 'other' can never match the type '.$@\data⏰!'
________________________________________________________________________________
```

This could be fixed by changing `ATOM = [a-zA-Z_\?@][a-zA-Z\._0-9\?@]+` to `ATOM = .+` in lexer, but I don't feel well with making such PR without proper unit tests

asummers:
@jeremyjh I'm going to tackle the multiple format problem (short, long, explain, flycheck). My plan is to take each error (`  defp message_to_string({:callback_info_missing, [behaviour]}) do` and throw it into a module, with a behaviour that defines what the message should look like for each format (`format_flycheck`, `format_short`, etc.). Any thoughts on that?

asummers:
Even better, we can divorce explain entirely from the context of a single error (maybe we can do better later, but at beginning this seems good) by saying something like `mix dialyxir --explain callback_info_missing`. This would allow all the format functions to just take in the arguments, because you can create a mapping of the name to validator, and then do lookup for each warning. 

What this ALSO accomplishes is the ability to easily do filtering by expressing something like:

```
path/to/file.ex:10: callback_info_missing 
```

or even something more exotic as 

```
path/to/file.ex: callback_info_missing 
```

and making that filter a little smarter can restore the old dialyzer warning by just running through the old formatter logic for purposes of knowing if you should be filtering it.

hassan:
Error, as requested: [from a very simple Phoenix app](https://gist.github.com/hassan/23688e44e50858028acb32804998791a)

asummers:
Hmmm a regression must have been introduced. I'll fix and add a test, then shoot you a ping when it's done. Thank you.

asummers:
@hassan Sorry about that. All fixed.

asummers:
@NobbZ Fixed!

NobbZ:
No, its not fixed:

```
** (FunctionClauseError) no function clause matching in Dialyxir.Warnings.PatternMatch.format_long/1

    The following arguments were given to Dialyxir.Warnings.PatternMatch.format_long/1:

        # 1
        ['pattern {\'ok\', Vuser@1}', '{\'error\',<<_:64,_:_*8>>}']

    Attempted function clauses (showing 1 out of 1):

        def format_long([[pattern, type]])

    lib/dialyxir/warnings/pattern_match.ex:10: Dialyxir.Warnings.PatternMatch.format_long/1
    lib/dialyxir/formatter.ex:123: Dialyxir.Formatter.format_warning/2
    lib/dialyxir/formatter.ex:72: anonymous fn/3 in Dialyxir.Formatter.format_and_filter/3
    (elixir) lib/enum.ex:1294: Enum."-map/2-lists^map/1-0-"/2
    lib/dialyxir/formatter.ex:69: Dialyxir.Formatter.format_and_filter/3
    lib/dialyxir/dialyzer.ex:15: Dialyxir.Dialyzer.Runner.run/2
    lib/dialyxir/dialyzer.ex:31: Dialyxir.Dialyzer.dialyze/3
    lib/mix/tasks/dialyzer.ex:203: Mix.Tasks.Dialyzer.run_dialyzer/2
    (mix) lib/mix/task.ex:314: Mix.Task.run_task/3
    (mix) lib/mix/cli.ex:80: Mix.CLI.run_task/2
    (elixir) lib/code.ex:677: Code.require_file/2
```

Now its worse than before, as I can not even ignore the warning any more (its caused by an upstream bug).

asummers:
Hmm. Sorry about that. Fixed part of your issue, not the whole thing apparently. I'm going to be working on the filtering stuff today so I'll make sure that ability gets restored.

asummers:
Now that just says `user`.

asummers:
@NobbZ the old Dialyzer filter rules should now work too. Existing rules that you already have in your ignore warnings should now be honored. I'm going to make that filtering process better for the new warnings. as well, but this should at least reduce some of the friction in the interim.

zachdaniel:
Can't post the whole error, but:
```
{:warn_matching, {'apps/my_app/lib/foo/bar/biz/buz.ex', 345}, {:pattern_match_cov, ['pattern <Vquery@1, _>', '<\'nil\' | \#{\'__struct__\':=\'Elixir.Ecto.Query\', _=>_},\
```

```
** (SyntaxError) nofile:1: syntax error before: '<'
    (elixir) lib/code.ex:442: Code.format_string!/2
    lib/dialyxir/pretty_print.ex:48: Dialyxir.PrettyPrint.pretty_print_type/1
    lib/dialyxir/warnings/pattern_match_covered.ex:12: Dialyxir.Warnings.PatternMatchCovered.format_long/1
    lib/dialyxir/formatter.ex:129: Dialyxir.Formatter.format_warning/2
    lib/dialyxir/formatter.ex:76: anonymous fn/3 in Dialyxir.Formatter.format_and_filter/3
```

NobbZ:
My spuervisor initialiser (simplified) produces the following warning:

```
lib/gitster/supervisor.ex:18
The call Supervisor.init(children :: [], opts :: [{:name, Gitster.Supervisor}, ...]) breaks the contract
([supervisor.child_spec() | {module(), term()} | module()], [init_option()]) :: {:ok, tuple()}
```

raw:

```
{:warn_failing_call, {'lib/gitster/supervisor.ex', 18}, {:call, [Supervisor, :init, '(Vchildren@1::[],Vopts@1::[{\'name\',\'Elixir.Gitster.Supervisor\'},...])', [2], :only_contract, '(maybe_improper_list(),[{atom(),_}])', '{\'ok\',{\#{\'intensity\':=_, \'period\':=_, \'strategy\':=_},[any()]}}', {true, '([supervisor:child_spec() | {module(),term()} | module()],[init_option()]) -> {\'ok\',tuple()}'}]}}
```

Code:

```
def init(:ok) do
  children = []
  opts = [name: __MODULE__]
  Supervisor.init(children, opts)
end
```

As you can see erlang modules are not prefixed by a colon in the pretty printed elixirified output.

asummers:
@NobbZ Fixed!

jeremyjh:
@asummers I agree it makes sense as a standalone library. The downside to that for `Dialyxir` would be that installing as a Global archive task does not support hex dependencies. I'm not sure if many people use it that way, but actually for the development of Dialyxir itself and the CI job - the only way I  could get Dialyxir to dialyze itself without random beam crashes was to install it as a global task in a separate mix environment. So if we externalize it I may have to use it as a submodule dependency rather than a hex dependency. Still, I'm not opposed to that and perhaps these issues can be solved another way.

asummers:
I'd be intrigued to tinker at that once this gets more stable, because that seems like it should be a solvable problem. We're a ways off from when I'd want to split it anyway, and it should be a clean break if so, so I'm not too worried about anything around that 😄

asummers:
My previous PR adds several helpful goodies.

Add `format_short` to all warnings which provides short one line messages for each error.  This should be useful for things like flycheck and other editor warnings. Similar to Credo's `--oneline`

Add explanations to MOST warnings, with a CTA for getting live examples for the remaining explanations as I couldn't figure out how to trip them. 

Adjusts warning format to include the error type (e.g. `no_return`) so that when you see an error like

```
lib/dialyxir/warnings/race_condition.ex:14:no_return
Function format_long/1 has no local return.
```

you can then know to say
```
mix dialyzer --explain no_return
```

and get an explanation about that class of error.

Add `--format` flag which takes in `--raw` `--short` `--dialyzer` `--dialyxir`. If omitted, defaults to `--dialyxir` behavior.

Also has some general error cleanup and housekeeping.

Some of the language could be better, but it seemed fine enough for a first pass.

/cc @jeremyjh

zachdaniel:
Would it be possible to explain all unique errors below a failed run? If not, a note at the bottom along the lines of "use this to explain the errors you've found" would be great.

asummers:
I've added support for internationalized atoms. This should also clear most of the unparsed atoms from structs/patterns/whatever across the board. 

I'm currently aware of a few unreported parsing bugs, but I think everything in this thread has been taken care of so far, as far as bugs. 

/cc @jeremyjh

jeremyjh:
@asummers yes I think you can pretty much check the "Test on real life input" box. How do you picture the "short errors" working. Will that need to compact the term representations somehow? For structs that could just be `%StructName{..}` but I don't see a general way to do it for other types. Or does this refer to another concern entirely?

asummers:
@jeremyjh I have already added the short errors with `--format short` (also added `--format raw` to replace `--raw`). Each warning has a `format_short/1` that takes in the same args as `format_long/1` and can selectively choose to pretty print things. Most of them just ignore everything but some in the interest of actually being useful (unknown behaviour e.g.) will print their MFAs and some other small info. Seems to work really well for me and never exceeds a single line.

The `explain` function (from `--explain no_return` or whatever warning code you wanna ask about) currently takes 0 arguments and has small examples to cause that error for as many of the ones I could figure out how to make happen in Elixir.

asummers:
@jeremyjh Okay at this point, I have eliminated all known bugs, added docs, fixed all tests. 

I am unsure what the Dialyzer errors are, and my recollection from previous investigation was a deeper area bubbling up. No matter, we can ignore them.

I have not added in the new filtering logic yet, but was wondering if you ever saw my proposal in:
https://github.com/jeremyjh/dialyxir/pull/118#issuecomment-385181523

jeremyjh:
@asummers Thanks for drawing my attention to that, I had missed it. I like what you've done with the warnings modules and explain functionality, it reminds me of the Credo architecture. You've put in an amazing amount of work on this. I agree what you suggest about providing filename/optionally line number and warning atom would be a good way to approach filtering. 

I think it might be nice to accept these as terms in an `.exs` file that we just do a `Code.eval_file` to load, this would leave the door open to supporting predicate filters. e.g.


```elixir
# .dialyxir.ignore.exs
[
{"filename.ex", :no_return},
{"other_filename.ex", 12, :no_return},
{"third_filename.ex", :no_return, fn _warning -> true end},
]
```

As for the travis issues - it looks like most of the warnings we have are regarding no local return, I think most of them go back to the fact that `pretty_print` can throw. One way to deal with these warnings, is update the spec to indicate `no_return()` - e.g. changing this silenced the warning on pretty_print for me: 

```elixir
  @spec pretty_print(String.t()) :: no_return()
  def pretty_print(str) do
```

Unfortunately, all the functions that call `pretty_print` need the same treatment.

The unknown function warnings can just be ignored.

We also need to remove `  - 1.5.3 ` from `.travis.yml` - addressing all those things should make travis appy.

asummers:
Curious what use you would see for the predicates functions. They're as easy to add as the others would be, but I'm not seeing the utility, immediately.

jeremyjh:
Well a predicate would you let you ignore something more specific than a file name and less transient than a line number, but to be honest I struggle to come up with an example that couldn't be solved with string matching. I do want to keep the simple string matching system as it is very easy for people to use for the most common cases.

asummers:
@jeremyjh How about the code I have just merged?

https://github.com/jeremyjh/dialyxir/pull/159

jeremyjh:
@asummers excellent, I really like the matching implementation!

asummers:
Great. Then I'm currently unaware of any particular tasks remaining on this branch.

.dialyzer_ignore.exs
```diff
@@ -0,0 +1,51 @@
...
+  {":0:unknown_type Unknown type: :wxCursor.wxCursor/0."},
+  {":0:unknown_type Unknown type: :wxDC.wxDC/0."},
+  {":0:unknown_type Unknown type: :wxFont.wxFont/0."},
+  {":0:unknown_type Unknown type: :wxMenu.wxMenu/0."},
+  {":0:unknown_type Unknown type: :wxWindow.wxWindow/0."},
+  {"lib/dialyxir/pretty_print.ex", :no_return, 100},
+  {"lib/dialyxir/pretty_print.ex", :no_return, 121},
+  {"lib/dialyxir/pretty_print.ex", :no_return, 69},
+  {"lib/dialyxir/pretty_print.ex", :no_return, 81},
+  {"lib/dialyxir/warning_helpers.ex", :no_return, 8},
```
jeremyjh: I would be fine just ignoring this warning in the entire warning module file - these files are trivial but if we add or delete text the line offsets might change right?

asummers: Fine by me.

lib/dialyxir/warning.ex
```diff
@@ -0,0 +1,40 @@
+defmodule Dialyxir.Warning do
+  @moduledoc """
+  Behaviour for defining warning semantings.
+
+  Contains callbacks for various warnings
+  """
+
+  @doc """
+  By expressiing the warning that is to be matched on, error handlong
```
jeremyjh: expressing spelling

lib/dialyxir/warning.ex
```diff
@@ -0,0 +1,40 @@
...
+  Explanation for a warning of this type. Should include a simple example of how to trigger it.
+  """
+  @callback explain() :: String.t()
+
+  @spec default_explain() :: String.t()
+  def default_explain() do
+    """
+    This warning type does not have an explanation yet. If you have
+    code that causes it, please file an issue or pull request in
+    https://github.com/jeremyjh/dialyxir/pull/118
```
jeremyjh: Don't want to forget to remove this PR reference before merging.

asummers:
@jeremyjh Merged 😄

zachdaniel:
How does the dialyzer ignore file play with these rules? Does the ignore have to match the original rule or the new output text? Or does either work?

asummers:
@zachdaniel Either/or, depending on your format. If you have old style ignores, they'll be honored, and if new, they will be honored, but the ignore can only point to one file so you have to pick.  I kinda assume that when you upgrade you'll use the new style text. But that could be better migrated.

andrewhao:
Just tried this out - bravo!

PR #120, opened 2018-04-26
Add named argument parser for specs.

asummers:
@jeremyjh Not sure what the best way to do this is. PRs into the branch here? Would be easier if I had write access, I think. 

Resolves https://github.com/jeremyjh/dialyxir/pull/118#issuecomment-384466365


jeremyjh:
This is fine but I'll add you as a contributor.

PR #127, opened 2018-04-27
Formatter support strings with length != 1

amatalai:
Minor improvement for issue with strings reported 2 days ago https://github.com/jeremyjh/dialyxir/pull/118#issuecomment-384486361

Now it should support empty strings as well as strings longer than 1 char.

Tested with 
```elixir
defmodule StringIssue2 do
  def single_arg do
    do_single_arg("")
  end

  def multiple_args do
    do_multiple_args("", 12345)
  end

  defp do_single_arg(var1) when is_atom(var1), do: :ok
  defp do_multiple_args(var1, _var2) when is_atom(var1), do: :ok
end

defmodule StringIssue3 do
  def single_arg do
    do_single_arg("String")
  end

  def multiple_args do
    do_multiple_args("String", 12345)
  end

  defp do_single_arg(var1) when is_atom(var1), do: :ok
  defp do_multiple_args(var1, _var2) when is_atom(var1), do: :ok
end
```

### DIFF
--- a/src/struct_parser.yrl
+++ b/src/struct_parser.yrl
@@ -84,6 +84,7 @@ function -> 'fun(' list '->' value ')' : {function, {args, '$2'}, {return, '$4'}
 
 contract -> list '->' value : {contract, {args, '$1'}, {return, '$3'}}.
 
+ascii_list -> '#' '{' '}' '#' : {ascii, []}.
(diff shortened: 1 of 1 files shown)


asummers:
Thank you 😄

PR #129, opened 2018-04-28
Fixed issue with elixir var versioning

amatalai:
Fixes issues with `Vvar@1`-like var names

Middle value from `{:atom, _, atom}` was probably line number from erlang AST, because atom from named_value wasn't unwrapped.


amatalai:
Relates to #118

asummers:
```
Middle value from {:atom, _, atom} was probably line number from erlang AST, because atom from named_value wasn't unwrapped.
```

Yep, that's what that was.

asummers:
Thank you 😃

PR #140, opened 2018-04-29
Resolve many of the dialyzer warnings

asummers:
Also fixed some formatting and added more of the Dialyxir formatter code to the Elixir `.formatter.exs`

Will soon be able to wait for builds to pass before merging.


asummers:
Relates to #118

PR #169, opened 2018-06-06
Better formatting errors that won't blow up the task.

asummers:
Will now give a better error that we can work with in error messages when the formatter fails.

### DIFF
--- a/lib/dialyxir/pretty_print.ex
+++ b/lib/dialyxir/pretty_print.ex
@@ -89,7 +89,7 @@ defmodule Dialyxir.PrettyPrint do
       :ok
     end
     """
-    |> Code.format_string!()
+    |> format()
     |> Enum.join("")
     |> String.trim_leading(prefix)
     |> String.trim_trailing(suffix)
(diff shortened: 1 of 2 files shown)


lib/dialyxir/pretty_print.ex
```diff
@@ -355,4 +355,13 @@ defmodule Dialyxir.PrettyPrint do
   end
 
   defp deatomize_char(char), do: char
+
+  defp format(code) do
+    try do
+      Code.format_string!(code)
+    rescue
+      _ ->
```
zachdaniel: Might be good to grab the exception/message and include that, for better bug reports.

asummers: I haven't found the errors from any of the lex/parse/pretty print/format steps that helpful while debugging, honestly. Printing the raw text of the input that failed like I do in the catch and printing that out lets me rerun with the exact same input and get the same error we'd be logging there, with a shorter bug report 😄 .

zachdaniel: :+1:

PR #189, opened 2018-06-13
Move explain to standalone task.

### LINKED ISSUE #179: Add explanations for warnings with default/missing explanations.
- [ ] `Dialyxir.Warnings.AppCall`
- [ ] `Dialyxir.Warnings.BinaryConstruction`
- [ ] `Dialyxir.Warnings.CallWithOpaque`
- [ ] `Dialyxir.Warnings.CallWithoutOpaque`
- [ ] `Dialyxir.Warnings.ContractDiff`
- [ ] `Dialyxir.Warnings.FuncionApplicationArguments`
- [ ] `Dialyxir.Warnings.ImproperListConstruction`
- [ ] `Dialyxir.Warnings.OpaqueEquality`
- [ ] `Dialyxir.Warnings.OpaqeGuard`
- [ ] `Dialyxir.Warnings.OpaqueMatch`
- [ ] `Dialyxir.Warnings.OpaqueNonequality`
- [ ] `Dialyxir.Warnings.OpaqueTypeTest`
- [ ] `Dialyxir.Warnings.RaceCondition`
- [ ] `Dialyxir.Warnings.RecordConstruction` (likely not possible)
- [ ] `Dialyxir.Warnings.RecordMatching`   (likely not possible)
- [ ] `Dialyxir.Warnings.UnknownBehaviour`
- [ ] `Dialyxir.Warnings.UnknownFunction`

jeremyjh:
resolves #179

### DIFF
--- a/lib/mix/tasks/dialyzer/explain.ex
+++ b/lib/mix/tasks/dialyzer/explain.ex
@@ -33,8 +33,14 @@ defmodule Mix.Tasks.Dialyzer.Explain do
 
   defp explanation_text(warning_name) do
     warning = String.to_atom(warning_name)
-    module = Map.get(Dialyxir.Warnings.warnings(), warning)
-    (module && module.explain()) || "Uknown warning named: #{warning_name}"
+
+    case Map.get(Dialyxir.Warnings.warnings(), warning) do
+      nil ->
+        "Unknown warning named: #{warning_name}"
+
+      module ->
+        module.explain()
+    end
   end
 
   defp list_warnings do
--- a/lib/mix/tasks/dialyzer/explain.ex
+++ b/lib/mix/tasks/dialyzer/explain.ex
@@ -1,0 +1,55 @@
+defmodule Mix.Tasks.Dialyzer.Explain do
+  @shortdoc "Display information about dialzer warnings."
+
+  @moduledoc """
+  This task provides background information about dialyzer warnings.
+  If invoked without any arguments it will list all warning atoms.
+  When invoked with the name of a particular warning, it will display
+  information regarding it.
+
+  ## Command line options
+
+  * `[warning]`       - display information regarding warning
+
+  ```
+  mix dialyzer.explain pattern_match
+  ```
+  """
+  use Mix.Task
+  alias Dialyxir.Output
+
+  def run(args) do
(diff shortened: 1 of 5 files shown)


lib/mix/tasks/dialyzer/explain.ex
```diff
@@ -0,0 +1,55 @@
...
+      {_, [], _} ->
+        list_warnings() |> Output.info()
+
+      _ ->
+        Mix.Task.run("help", ["dialyzer.explain"])
+    end
+  end
+
+  defp explanation_text(warning_name) do
+    warning = String.to_atom(warning_name)
```
asummers: I'm always a bit weary of `String.to_atom/1`. Should be fine here, though, I think.

lib/mix/tasks/dialyzer/explain.ex
```diff
@@ -0,0 +1,55 @@
...
+    module = Map.get(Dialyxir.Warnings.warnings(), warning)
+    (module && module.explain()) || "Uknown warning named: #{warning_name}"
```
asummers: `explain/1` will always return a legit value, so this check is really 

```
if module do
  module.explain()
else
  "Unknown warning named: #{warning_name}"
end
```

Also typo on `Unknown`

asummers: Or if you REALLY want to get fancy:

```
case Map.get(Dialyxir.Warnings.warnings(), warning) do
  nil -> 
    "Unknown warning named: #{warning_name}"
  module ->
    module.explain()
end
```

jeremyjh: If the warning name isn't found in the map, `module` will be nil so the left branch will fail and it prints the unknown. Do you just mean the `if`/`case` is more readable?

asummers: Yeah I think it's a little more honest as to intent. They're all functionally the same.

lib/mix/tasks/dialyzer/explain.ex
jeremyjh: Yeah its the best choice here, we can't leak.

asummers:
Might be worth mentioning the new task in the README too

PR #201, opened 2018-06-24
Remove explain opt to prevent error.

### LINKED ISSUE #200: Unknown dialyzer command line option on Elixir 1.7-dev
When running on the local environment I get this error:
```
:dialyzer.run error: Unknown dialyzer command line option: {explain,nil}
```

Version RC2

Elixir/Erlang:
```
Erlang/OTP 20 [erts-9.3.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.7.0-dev (094eb6a) (compiled with Erlang/OTP 20)
```

asummers:
Resolves #200

### DIFF
--- a/lib/dialyxir/dialyzer.ex
+++ b/lib/dialyxir/dialyzer.ex
@@ -1,5 +1,5 @@
 defmodule Dialyxir.Dialyzer do
-  import Dialyxir.Output, only: [color: 2]
+  import Dialyxir.Output, only: [color: 2, info: 1]
   alias String.Chars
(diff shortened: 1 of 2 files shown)


jeremyjh:
@asummers I pushed another commit, we need to remove `list` as well, and changed where we log args as this was really intended to log the args we are actually passing to dialyzer.

asummers:
LGTM!

PR #204, opened 2018-06-27
Docs: use `--format dialyzer` to create ignore files

harlantwood:
When running the current master branch, `mix dialyzer` now creates output (while wonderful for humans!) which does not work when pasting (a single line of) it into the ignore file.  

It works fine if I use `mix dialyzer --format dialyzer` -- so this PR updates the docs to reflect that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,9 @@ end
 
 Any line of dialyzer output (partially) matching a line in `"dialyzer.ignore-warnings"` is filtered.
 
-For example, in project where `mix dialyzer` outputs:
+Note that copying output in the default format will not work!  Run `mix dialyzer --format dialyzer` to produce output suitable for the ignore file.
+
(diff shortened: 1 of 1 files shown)


asummers:
There's also the new style ignore discussed at the bottom of the README. Might be worth calling attention to that section in the copy, I think, for completeness.

jeremyjh:
I may tweak this a bit in my next PR per @asummers suggestion, thanks for calling attention to this.

harlantwood:
Certainly.  Thanks for the great work on this tool.  My team is really excited about the new formatting features; it's been a bit of a nightmare to wade through multiple screens of red erlang terms.  I actually was formatting them manually in a text editor when it got too bad. 

;)

asummers:
That was how it started for me :smile: Then I got myself into the deep end with lexers and parsers and all the absurd quirks of Dialyzer :stuck_out_tongue:

Issue #220, opened 2018-07-11
Warning Language Audit


asummers:
We should give each message a once over, particularly with respect to short message, so that we know that they make sense in real life.

- [x] `Dialyxir.Warnings.AppCall`
- [x] `Dialyxir.Warnings.Apply`
- [x] `Dialyxir.Warnings.BinaryConstruction`
- [x] `Dialyxir.Warnings.Call`
- [x] `Dialyxir.Warnings.CallToMissingFunction`
- [x] `Dialyxir.Warnings.CallWithOpaque`
- [x] `Dialyxir.Warnings.CallWithoutOpaque`
- [x] `Dialyxir.Warnings.CallbackArgumentTypeMismatch`
- [x] `Dialyxir.Warnings.CallbackInfoMissing`
- [x] `Dialyxir.Warnings.CallbackMissing`
- [x] `Dialyxir.Warnings.CallbackSpecArgumentTypeMismatch`
- [x] `Dialyxir.Warnings.CallbackSpecTypeMismatch`
- [x] `Dialyxir.Warnings.CallbackTypeMismatch`
- [x] `Dialyxir.Warnings.ContractDiff`
- [x] `Dialyxir.Warnings.ContractSubtype`
- [x] `Dialyxir.Warnings.ContractSupertype`
- [x] `Dialyxir.Warnings.ContractWithOpaque`
- [x] `Dialyxir.Warnings.ExactEquality`
- [x] `Dialyxir.Warnings.ExtraRange`
- [x] `Dialyxir.Warnings.FuncionApplicationArguments`
- [x] `Dialyxir.Warnings.FunctionApplicationNoFunction`
- [x] `Dialyxir.Warnings.GuardFail`
- [x] `Dialyxir.Warnings.GuardFailPattern`
- [x] `Dialyxir.Warnings.ImproperListConstruction`
- [x] `Dialyxir.Warnings.InvalidContract`
- [x] `Dialyxir.Warnings.MissingRange`
- [x] `Dialyxir.Warnings.NegativeGuardFail`
- [x] `Dialyxir.Warnings.NoReturn`
- [x] `Dialyxir.Warnings.OpaqueGuard`
- [x] `Dialyxir.Warnings.OpaqueEquality`
- [x] `Dialyxir.Warnings.OpaqueMatch`
- [x] `Dialyxir.Warnings.OpaqueNonequality`
- [x] `Dialyxir.Warnings.OpaqueTypeTest`
[...]

PR #222, opened 2018-07-13
Short Format Warning Updates

jeremyjh:
We'll use this PR to comment on code and or make agreed changes that come from review of https://github.com/jeremyjh/dialyxir/issues/220. I pushed an `app_call` suggestion mostly so we'd have a diff to open this PR.

I'd prefer to focus on short warning as longer form and explain will not break ignore file compatibility.

### DIFF
--- a/lib/dialyxir/warning_helpers.ex
+++ b/lib/dialyxir/warning_helpers.ex
@@ -83,4 +83,17 @@ defmodule Dialyxir.WarningHelpers do
   def form_position_string(arg_positions) do
     Enum.map_join(arg_positions, " and ", &ordinal/1)
   end
+
+  def unqualify_module(name) when is_binary(name) do
+    case String.split(name, ".") do
+      [only] -> only
+      multiple ->
+        multiple
+        |> Enum.reverse()
+        |> Enum.take(2)
+        |> Enum.reverse()
+        |> Enum.join(".")
+    end
+  end
+
 end
--- a/lib/dialyxir/warning_helpers.ex
+++ b/lib/dialyxir/warning_helpers.ex
@@ -86,14 +86,13 @@ defmodule Dialyxir.WarningHelpers do
 
   def unqualify_module(name) when is_binary(name) do
     case String.split(name, ".") do
-      [only] -> only
+      [only] ->
+        only
+
       multiple ->
         multiple
-        |> Enum.reverse()
-        |> Enum.take(2)
-        |> Enum.reverse()
+        |> Enum.take(-2)
         |> Enum.join(".")
     end
   end
-
 end
--- a/lib/dialyxir/warning_helpers.ex
+++ b/lib/dialyxir/warning_helpers.ex
@@ -83,16 +83,4 @@ defmodule Dialyxir.WarningHelpers do
   def form_position_string(arg_positions) do
     Enum.map_join(arg_positions, " and ", &ordinal/1)
   end
-
-  def unqualify_module(name) when is_binary(name) do
-    case String.split(name, ".") do
-      [only] ->
-        only
-
-      multiple ->
-        multiple
-        |> Enum.take(-2)
-        |> Enum.join(".")
-    end
-  end
 end
--- a/lib/dialyxir/warnings/app_call.ex
+++ b/lib/dialyxir/warnings/app_call.ex
@@ -7,10 +7,9 @@ defmodule Dialyxir.Warnings.AppCall do
 
   @impl Dialyxir.Warning
   @spec format_short([String.t()]) :: String.t()
-  def format_short([module, function, arity, _culprit, _expected_type, _actual_type]) do
+  def format_short([module, function, _arity, _culprit, _expected_type, _actual_type]) do
     pretty_module = Erlex.pretty_print(module)
-    "Module or function to apply is not an atom: #{pretty_module}.#{function}/#{arity}."
-    "#{pretty_culprit} in #{pretty_module}.#{function} is not an atom."
+    "Module or function to apply is not an atom: #{pretty_module}.#{function}."
   end
 
   @impl Dialyxir.Warning
--- a/lib/dialyxir/warnings/app_call.ex
+++ b/lib/dialyxir/warnings/app_call.ex
@@ -9,7 +9,7 @@ defmodule Dialyxir.Warnings.AppCall do
   @spec format_short([String.t()]) :: String.t()
   def format_short([module, function, arity, _culprit, _expected_type, _actual_type]) do
     pretty_module = Erlex.pretty_print(module)
-    "The call #{pretty_module}.#{function}/#{arity} has a type mismatch."
+    "Module or function to apply is not an atom: #{pretty_module}.#{function}/#{arity}."
   end
 
   @impl Dialyxir.Warning
--- a/lib/dialyxir/warnings/app_call.ex
+++ b/lib/dialyxir/warnings/app_call.ex
@@ -7,9 +7,8 @@ defmodule Dialyxir.Warnings.AppCall do
 
   @impl Dialyxir.Warning
   @spec format_short([String.t()]) :: String.t()
-  def format_short([module, function, _arity, _culprit, _expected_type, _actual_type]) do
-    pretty_module = Erlex.pretty_print(module)
-    "Module or function to apply is not an atom: #{pretty_module}.#{function}."
+  def format_short([_module, function | _]) do
+    "Module or function to apply is not an atom in #{function}."
   end
 
   @impl Dialyxir.Warning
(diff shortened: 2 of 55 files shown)


lib/dialyxir/warnings/app_call.ex
```diff
@@ -9,7 +9,7 @@ defmodule Dialyxir.Warnings.AppCall do
   @spec format_short([String.t()]) :: String.t()
   def format_short([module, function, arity, _culprit, _expected_type, _actual_type]) do
     pretty_module = Dialyxir.PrettyPrint.pretty_print(module)
-    "The call #{pretty_module}.#{function}/#{arity} has a type mismatch."
+    "Module or function to apply is not an atom: #{pretty_module}.#{function}/#{arity}."
```
jeremyjh: My understanding of `app_call` is based on the upstream [test](https://github.com/erlang/otp/blob/1526eaead833b3bdcd3555a12e2af62c359e7868/lib/dialyzer/test/small_SUITE_data/src/app_call.erl) and [results](https://github.com/erlang/otp/blob/1526eaead833b3bdcd3555a12e2af62c359e7868/lib/dialyzer/test/small_SUITE_data/results/app_call). The type mismatch refers to the module and function passed to apply or joined in function application syntax, versus args/arity. Module and function should always be atoms.

lib/dialyxir/warnings/call_with_opaque.ex
```diff
@@ -8,7 +8,7 @@ defmodule Dialyxir.Warnings.CallWithOpaque do
   @impl Dialyxir.Warning
   @spec format_short([String.t()]) :: String.t()
   def format_short(_) do
-    "Call with opaqueness type mismatch."
+    "Call with opaque type mismatch."
```
asummers: Maybe `"Type mismatch in call with opaque."`?

lib/dialyxir/warnings/app_call.ex
asummers: Yeah the tests look like it does. We should match on that and expand it out to the whole word. Not sure if that's a string or atom. Could probably just rig up their test suite in the explanation at followup for the other ticket.

asummers: Oh, no, it's the actual invocation, I think. So that'd be an atom. So like for us it'd be 
```
def frobnicate() do
  module = 42
  function = 123
  module.function()
end
```

And we'd get culprit as `module` there, which I think we'd wanna just call pretty print on.

lib/dialyxir/warnings/call_with_opaque.ex
asummers: Same verbiage for the `without` one. I'd say.

lib/dialyxir/warnings/app_call.ex
jeremyjh: I don't know if this error is possible in idiomatic Elixir; the way function application is rewritten we end up with a pattern match error on that snippet of code, or on one more like the Erlang example. Probably not worth a lot more time but I did pp the culprit.

jeremyjh:
Hey sorry I've been stalled on this for so long. I'll try to finish up by Tuesday night so we can include it in a hopefully finally RC.

asummers:
I've definitely stepped away from this for a bit -- @jeremyjh wanna get this going again so we can finalize the release? Shouldn't be too much work left to split.

jeremyjh:
@asummers I think I'm ok with most of the warnings as they are. Actually, I started adding culprit information to a lot of them but I think that defeats the purpose of a short warning. But yeah I will finish going through them.

jeremyjh:
ping @asummers feedback on above? I think I'd like to add one culprit - whatever would be most specific - to every short warning.

jeremyjh:
@asummers I wasn't imagining the user would learn about how the text is formatted - they would just copy the short error message they see emitted by dialyxir and put it in the ignore file. But I realized, this won't really accomplish my goal by itself since the line number is part of the short description. Maybe we would also need to support regex in the ignore file.

asummers:
Didn't mean to suggest they'd _have_ to, but being able to predict what you're going to have to do is a really nice property of development tools. If I have a long function, I know I'll probably have to disable the complexity checks in Credo, for example. Knowing what kind of ignore I'm going to have to add without the tool telling me might be an unrealistic goal, though, given the esoteric nature of some of the source material from Dialyzer. 

Re: regex, how would you see that working? Would you be able to mix and match regex and exact matches in the ignore file?

jeremyjh:
Yes regex would just be another match clause anywhere we current allow a binary match on message, we'd have another clause matching `%Regex{}` where we apply the regex.

In master we already have culprit information in a few short error messages. I think what most people would expect, is that an error contains the most descriptive relevant identifier for the error in question. As far as I know Credo only supports source annotations in comments that ignore an entire file, or ignore the next `n` lines - so as your function grows you have to update the annotation. I don't think that is the right model. 

If you have another suggestion for how we can selectively ignore a particular instance of an error without relying on line numbers I'm certainly open to it.

jeremyjh:
@asummers I've started implementing the changes based on my comment on the 18th, in this [commit](https://github.com/jeremyjh/dialyxir/pull/222/commits/5783e91f57be6e8ce5aa99e277a7cab12232e946). Basically, I try to include exactly one piece of identifying information in the warning text where available.

I also started putting some of your examples into permanent fixtures so you can see the warnings, so you can now run `MIX_ENV=examples mix dialyzer --format short` to see some of these rendered.

jeremyjh:
@asummers I think I'm done with this, do you want to review? You will see a lot of examples rendered if you run `MIX_ENV=examples mix dialyzer --format short`. I pretty much only extracted examples in cases where we already had them in the explain, so not everything is covered there, but quite a bit is.

test/examples/unknown_type.ex
```diff
@@ -0,0 +1,9 @@
+defmodule UnknownTypeExample do
```
asummers: These modules should probably all be named according to their directory/file name.

lib/dialyxir/warnings/app_call.ex
```diff
@@ -7,9 +7,8 @@ defmodule Dialyxir.Warnings.AppCall do
 
   @impl Dialyxir.Warning
   @spec format_short([String.t()]) :: String.t()
-  def format_short([module, function, arity, _culprit, _expected_type, _actual_type]) do
-    pretty_module = Erlex.pretty_print(module)
-    "The call #{pretty_module}.#{function}/#{arity} has a type mismatch."
+  def format_short([_module, function | _]) do
+    "Module or function to apply is not an atom in #{function}."
```
asummers: This verbiage is kind of suboptimal, I think. I don't have any explicit suggestions offhand, but I would have no idea what to do with that output.

lib/dialyxir/warning_helpers.ex
```diff
@@ -83,4 +83,16 @@ defmodule Dialyxir.WarningHelpers do
   def form_position_string(arg_positions) do
     Enum.map_join(arg_positions, " and ", &ordinal/1)
   end
+
+  def unqualify_module(name) when is_binary(name) do
```
asummers: Not a huge deal, but is this something that should live next to the one place using it as a `defp`, instead?

lib/dialyxir/warnings/callback_argument_type_mismatch.ex
```diff
@@ -7,12 +7,8 @@ defmodule Dialyxir.Warnings.CallbackArgumentTypeMismatch do
...
   @impl Dialyxir.Warning
   @spec format_short([String.t()]) :: String.t()
-  def format_short([behaviour, function, arity, position, _success_type, _callback_type]) do
-    pretty_behaviour = Erlex.pretty_print(behaviour)
-    ordinal_position = Dialyxir.WarningHelpers.ordinal(position)
-
-    "Type mismatch in #{ordinal_position} argument for #{function}/#{arity}" <>
-      " callback in #{pretty_behaviour} behaviour."
+  def format_short([_behaviour, function | _]) do
+    "Type mismatch for argument to callback #{function}."
```
asummers: We probably want arity here, too. Maybe something like:

```
def format_short([_behaviour, function, arity | _]) do   
  "Type mismatch with behaviour callback to #{function}/#{arity}."
end
```

lib/dialyxir/warnings/extra_range.ex
```diff
@@ -7,11 +7,8 @@ defmodule Dialyxir.Warnings.ExtraRange do
 
   @impl Dialyxir.Warning
   @spec format_short([String.t()]) :: String.t()
-  def format_short([module, function, arity, _extra_ranges, _signature_range]) do
-    pretty_module = Erlex.pretty_print(module)
-
-    "@spec for #{pretty_module}.#{function}/#{arity} has more types " <>
-      "than returned by function."
+  def format_short([_module, function | _]) do
+    "@spec for #{function} has more types " <> "than returned by function."
```
asummers: We can drop the `<>` here

lib/dialyxir/warnings/app_call.ex
jeremyjh: I'm not sure its possible to trigger this warning in Elixir. But my understanding is it would be something along the lines of `apply(__MODULE__, "not an atom", [])` but that gets you `no_return` and `call` warnings.

PR #225, opened 2018-07-22
Extract pretty printer to separate library.

asummers:
This will allow parallel work for e.g. the GSOC project, Wobserver, SASL logs, and others.


asummers:
You had mentioned at one point about running into issues with it Dialyzing itself. Curious what those were, this was mostly straightforward, and I didn't seem to run into them.

mix.exs
```diff
@@ -24,7 +24,7 @@ defmodule Dialyxir.Mixfile do
   end
 
   def application do
-    [mod: {Dialyxir, []}, applications: [:dialyzer, :crypto, :mix]]
+    [mod: {Dialyxir, []}, applications: [:dialyzer, :crypto, :mix, :erlex]]
```
jeremyjh: I think this is redundant; since 1.4 Elixir adds all deps to the application list unless they have `runtime: false`.

jeremyjh:
Well the issue with dialyzing itself doesn't happen when you install the task as a global archive. I didn't expect to be able to continue doing that since archives do not include dependencies. I'm guessing it still loads Erlex from the local dependency in this case, so it continues to work for us. We will need to remove the instructions about installing `dialyxir` globally from the README though as that will not continue to work for people.

asummers:
Hmmm I wonder if we should just keep those and just clone the steps to point to my repo too. Then people would continue to be able to use that strategy.

mix.exs
asummers: I'll play with it and see what all I can purge from here.

asummers:
Credo does something similar with having you install Bunt, so it's not completely without precedence. 

https://github.com/rrrene/credo#using-credo-as-stand-alone

mix.exs
```diff
@@ -24,7 +24,7 @@ defmodule Dialyxir.Mixfile do
   end
 
   def application do
-    [mod: {Dialyxir, []}, applications: [:dialyzer, :crypto, :mix, :erlex]]
```
jeremyjh: I actually just meant the `:erlex` was redundant - the other apps should be there since they are not `deps`.

asummers: It seemed to start okay and didn't freak out when used.

jeremyjh: Yeah I don't think any of those actually depend on getting a start as they are today,  but I think its best to leave them in terms of OTP correctness.

asummers: Did some research and apparently that's supposed to be `extra_applications` for things like crypto.

asummers:
@jeremyjh Gonna go ahead and merge this I think. It now has `extra_applications` where it used to have `applications`

jeremyjh:
> It now has extra_applications where it used to have applications

Excellent, good catch.

Issue #235, opened 2018-08-21
Unknown warning: contract range


evnu:
I encountered the following unknown warning with dialyxir 1.0-rc3, which cannot be pretty-printed yet:

```
Please file a bug in https://github.com/jeremyjh/dialyxir/issues with this message.

Unknown warning:
:contract_range

Legacy warning:
apps/utils/lib/my_module.ex:103: The contract 'Elixir.MyModule':v1(timestamp::'Elixir.DateTime':t() | 'Elixir.Unixtime':t(),mac_address::<<_:136>>,clock_seq::<<_:14>>,format::'binary' | 'string') -> {'ok',uuid::t() | t_string()} | {'error',atom()}
    ; (timestamp::<<_:60>>,mac_address::'Elixir.MAC':t_bin(),clock_seq::<<_:14>>,format::'binary' | 'string') -> t() | t_string() cannot be right because the inferred return for v1(<<_:60>>,_mac_binary@1::<<_:48>>,<<_:14>>,_format@1::any()) on line 130 is {'error','wrong_argument'} | {'ok',<<_:128,_:_*160>>}
```

asummers:
Huh interesting. Thought I had that one. I'll get it added soon.

asummers:
@evnu is there a public project I can see this on by chance? Want to make sure my fix reads okay.

evnu:
No, I don't have a public project with this issue. Testing on a branch sounds reasonable, maybe I can also find a small code snippet to produce the error.

evnu:
Finding a small example is surprisingly complicated. I found an Erlang example on [stackoverflow](https://stackoverflow.com/a/11899970/436853), which produces this warning, but cannot get this to produce the same warning after converting it to Elixir.

asummers:
No trouble! I have had that issue as well with certain errors. I'll push a branch and we can go from there.

asummers:
https://github.com/jeremyjh/dialyxir/tree/contract-range-warning Here's the branch! Let me know if it looks okay for you. Should just be able to plug and play.

evnu:
Looks perfect to me:

```
apps/utils/lib/my_module.ex:103:contract_range
Contract cannot be correct because return type on line number 130 is mismatched.

Function:
MyModule.v1(<<_::60>>, _mac_binary :: <<_::48>>, <<_::14>>, _format :: any())

Type specification:
Contract head: (
  timestamp :: DateTime.t() | Unixtime.t(),
[...]

PR #236, opened 2018-08-22
Add :contract_range warning

asummers:
Also bump Erlex to add newlines to multi contract head warnings. 

Resolves https://github.com/jeremyjh/dialyxir/issues/235

Example output: 
```
apps/utils/lib/my_module.ex:103:contract_range
Contract cannot be correct because return type on line number 130 is mismatched.

Function:
MyModule.v1(<<_::60>>, _mac_binary :: <<_::48>>, <<_::14>>, _format :: any())

Type specification:
Contract head: (
  timestamp :: DateTime.t() | Unixtime.t(),
  mac_address :: <<_::136>>,
  clock_seq :: <<_::14>>,
  format :: :binary | :string
) :: {:ok, uuid :: t() | t_string()} | {:error, atom()}
Contract head: (
  timestamp :: <<_::60>>,
  mac_address :: MAC.t_bin(),
  clock_seq :: <<_::14>>,
  format :: :binary | :string
) :: uuid :: t() | t_string()

Success typing (line 130):
{:error, :wrong_argument} | {:ok, <<_::128, _::size(160)>>}
```


asummers:
Will investigate Dialyzer failure.

jeremyjh:
You can just ignore that warning, I think its the same we have in every warnings module, I think pretty_print or one of the functions it calls can throw, making our @spec inaccurate.

Issue #244, opened 2018-09-13
Parse error


ConnorRigby:
```
lib/nerves_time/ntpd.ex:69:pattern_match_cov
The pattern
:variable_

can never match since previous clauses completely cover the type
nil
________________________________________________________________________________
Please file a bug in https://github.com/jeremyjh/dialyxir/issues with this message.

Failed to parse warning:
[{:"(", 1}, {:atom_part, 1, 'P'}, {:atom_part, 1, 'o'}, {:atom_part, 1, 'r'}, {:atom_part, 1, 't'}, {:atom_part, 1, 'N'}, {:atom_part, 1, 'a'}, {:atom_part, 1, 'm'}, {:atom_part, 1, 'e'}, {:",", 1}, {:atom_part, 1, 'P'}, {:atom_part, 1, 'o'}, {:atom_part, 1, 'r'}, {:atom_part, 1, 't'}, {:atom_part, 1, 'S'}, {:atom_part, 1, 'e'}, {:atom_part, 1, 't'}, {:atom_part, 1, 't'}, {:atom_part, 1, 'i'}, {:atom_part, 1, 'n'}, {:atom_part, 1, 'g'}, {:atom_part, 1, 's'}, {:")", 1}, {:->, 1}, {:atom_part, 1, 'p'}, {:atom_part, 1, 'o'}, {:atom_part, 1, 'r'}, {:atom_part, 1, 't'}, {:"(", 1}, {:")", 1}, {:atom_part, 1, 'w'}, {:atom_part, 1, 'h'}, {:atom_part, 1, 'e'}, {:atom_part, 1, 'n'}, {:atom_part, 1, 'P'}, {:atom_part, 1, 'o'}, {:atom_part, 1, 'r'}, {:atom_part, 1, 't'}, {:atom_part, 1, 'N'}, {:atom_part, 1, 'a'}, {:atom_part, 1, 'm'}, {:atom_part, 1, 'e'}, {:::, 1}, {:"{", 1}, {:atom_full, 1, '\'spawn\''}, {:",", 1}, {:atom_part, 1, 'C'}, {:atom_part, 1, ...}, {:atom_part, ...}, {...}, ...]


Legacy warning:
[...]

PR #246, opened 2018-09-18
Bump Erlex to 0.1.4

asummers:
Adds `when` support in contract heads

Resolves https://github.com/asummers/erlex/pull/16
Resolves https://github.com/jeremyjh/dialyxir/issues/244


asummers:
General aside, this was _way_ simpler than I was expecting.

PR #249, opened 2018-09-19
Add opaque_match explanation and clean up format long/short.

asummers:
While debugging I also noticed that the `form_position_string` should be pretty printing the ordinals in its output so I tripped that too. More errors will be forthcoming as I figure out how to trip more opaque things.


asummers:
Hmm how annoying. It doesn't like the 

```elixir
throw({:error, :formatting, code})
```

from Erlex, which we explicitly catch.  Considering ignoring that error for now, til I can figure out how to make Erlex's Dialyzer all pass.

asummers:
Ah yeah we do that for all the others.

Issue #227, opened 2018-07-23
Updating PTL for dependency changes


micahrye:
I am using Dialyxir in my projects and have run into an issue with changes in local dependencies not being recognized by dialyxir when running `mix dialyzer`. 

I my project I have local dependencies that get referenced in the parent applications `mix.exs` file via `path:` 

When making changes to the dependency the dialyzer plt is never updated. 

My understanding, Erlang dialyzer, is that the command line argument [--check_plt](http://erlang.org/doc/man/dialyzer.html) should result in dialyzer checking "the PLT for consistency and rebuild it if it is not up-to-date." 

`dialyxir`  does not support the cmd line argument `--check_plt`. 

I modified my the `project/0` function of `mix.exs` and added:

```
[
...
dialyzer: [ check_plt: true ]
...
]
```
Thinking that I could pass arguments this way, but that also does not work. 

**I would like** that when I make changes to a dependency that those changes are recognized and the `plt` is updated without having to recompile the entire `plt`.

jeremyjh:
We hash `mix.lock` and check the PLT when it changes. For local dependencies, we could add an option to force checking all the time. Its not fast, even when there are no changes. One other option would be to add the paths via `paths` key, rather than as a dependency. That would be slower too, since it would actually check the dependency as as well.

micahrye:
Thanks @jeremyjh, is there a "convenient" way for me to update the PLT from the cmd line or other args? 

Your response seemed to be talking about potential approaches to handle the issue of local dependencies.

micahrye:
As I have time I will hack about it some and see if I can put together anything nice. If so I will put a PR out. 

First jump in to force it to update PLT is slow, but faster than full rebuild. 

Ideally, it is not often that you need to recompile the PLT even for a single dependency, so even if it is a bit slow once that is not soooo bad, as long as it is still much faster than recompile of the entire PLT. 

Suggestions on approach appreciated. 

thx.

jeremyjh:
The way to do this would be to add a new argument `--check-plt`, and pass that down to https://github.com/jeremyjh/dialyxir/blob/273156eecf597c641567dca0a86ff79e846a8e81/lib/mix/tasks/dialyzer.ex#L203 to skip checking the dependency_hash when it is true.

mguimas:
[...]

PR #250, opened 2018-09-24
Add flag for forcing PLT check

albertored:
When dealing with local dependencies the generated PLT is not checked upon changing of of them.
Now passing the flag `--force-check` the PLT is checked even if the lock file hash is not changed.

See #227

### DIFF
--- a/lib/mix/tasks/dialyzer.ex
+++ b/lib/mix/tasks/dialyzer.ex
@@ -10,6 +10,8 @@ defmodule Mix.Tasks.Dialyzer do
 
     * `--no-compile`       - do not compile even if needed.
     * `--no-check`         - do not perform (quick) check to see if PLT needs updated.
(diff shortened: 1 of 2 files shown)


albertored:
@asummers thanks, changes applied

lib/mix/tasks/dialyzer.ex
```diff
@@ -214,7 +218,7 @@ defmodule Mix.Tasks.Dialyzer do
 
   defp run_dialyzer(opts, dargs) do
     args = [
-      {:check_plt, false},
+      {:check_plt, opts[:force_check] || false},
```
asummers: Is the `|| false` necessary here anymore because of the Keyword.get?

asummers:
@jeremyjh Any issues with this? Gonna update and merge if not.

PR #254, opened 2018-10-10
Fix time calculation to do us to min/sec properly.

asummers:
This has been bothering me, so I'm fixing it.

### DIFF
--- a/lib/dialyxir/formatter.ex
+++ b/lib/dialyxir/formatter.ex
@@ -8,8 +8,8 @@ defmodule Dialyxir.Formatter do
   import Dialyxir.Output, only: [info: 1]
 
   def formatted_time(duration_ms) do
-    minutes = div(duration_ms, 6_000_000)
-    seconds = (rem(duration_ms, 6_000_000) / 1_000_000) |> Float.round(2)
+    minutes = div(duration_ms, 60_000)
+    seconds = (rem(duration_ms, 60_000) / 1_000) |> Float.round(2)
     "done in #{minutes}m#{seconds}s"
   end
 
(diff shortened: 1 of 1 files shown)


jeremyjh:
It looks right to me as it is.  [tc](http://erlang.org/doc/man/timer.html#tc-1) reports time in *microseconds*.  If there is anything wrong presently, it is that the argument name is `duration_ms` instead of `duration_us`

On master if I `/usr/bin/time mix dialyzer` I get: 

```
done in 0m1.28s
done (passed successfully)
        1.87 real         2.99 user         0.37 sys
```

asummers:
Locally I'm getting it reporting several minute times when my console and wall clock are both reporting 20 seconds or so.

asummers:
Oh I see what's wrong. It should be `60_000_000` not `6_000_000`

jeremyjh:
Now this branch is reporting `done in 0m3607.16s` in travis, while travis reports that step took 3 minutes. Which OTP/Elixir are you using locally? I get the right answers on OTP 21/1.7.3.

jeremyjh:
@asummers sorry, my comment was mistaken. master only works on projects that take less than six seconds to analyze. This is sad, but I've believed for months our project at work takes a minute to analyze because I can't tell the subjective difference between 10s and 1m. 

Still, the second line you changed in the PR should have the denominator reverted as the seconds value is incorrectly reporting 3607 seconds instead of 3.61; rem returns a time in microseconds so it must be divided by 1_000_000. Then travis should report a time of ~3s, which is correct since its only counting the time doing analysis, not the PLT step.

PR #256, opened 2018-10-16
Regex Ignores

### LINKED ISSUE #210: Provide a way to ignore folder from elixir-ish ignore file
Following up discussion on Slack it's sometimes desired to simply ignore warnings produced by tests and ignoring all warnings from `test/` folder would work for that.

With dialyzer ignore format you can just type a subpath: `/test/` and it would silence all warnings that mention that path.

jeremyjh:
resolves #210 

Implements my suggestion in a comment on #222 about supporting regex ignores. These work by embedding a literal `Regex` in the ignore file's list, which is matched against the short_description. This makes it possible for ignores to be very specific without being tied to a single line number, and also more general. This makes it possible to cut down Dialyxir's own ignore file quite substantially.


jeremyjh:
@asummers do you want to take a look at this?

asummers:
I'm never going to use this in my own projects, but this looks reasonable enough.

Issue #260, opened 2018-10-22
Dialyxir skips errors with blank ignore file


jswny:
When adding `dialyzer: [ignore_warnings: ".dialyzer_ignore.exs"]` to my `mix.exs`, and then a `.dialyzer_ignore.exs` file with nothing but an empty list (`[]`) Dialyxir ignores certain errors.

I do not think this is intended but if it is please correct my usage of the library.

More specifically, without the `dialyzer: [ignore_warnings: ".dialyzer_ignore.exs"]` line, running `mix dialyzer` gives the following output:
```
warning: found quoted keyword "test" but the quotes are not required. Note that keywords are always atoms, even when quoted, and quotes should only be used to introduce keywords with foreign characters in them
  mix.exs:68

Compiling 45 files (.ex)
Generated aau_stats app
Finding suitable PLTs
Checking PLT...
[:compiler, :connection, :cowboy, :cowlib, :crypto, :db_connection, :decimal, :ecto, :eex, :elixir, :gettext, :kernel, :logger, :mime, :phoenix, :phoenix_ecto, :phoenix_html, :phoenix_pubsub, :plug, :poison, :poolboy, :postgrex, :ranch, :stdlib]
PLT is up to date!
Starting Dialyzer
[
  check_plt: false,
  init_plt: '/Users/joe/GitHub/aau_stats/_build/dev/dialyxir_erlang-21.0_elixir-1.7.1_deps-dev.plt',
  files_rec: ['/Users/joe/GitHub/aau_stats/_build/dev/lib/aau_stats/ebin'],
  warnings: [:unknown]
]
Total errors: 3, Skipped: 0
done in 0m2.22s
:0:unknown_function
Function Phoenix.LiveReloader.call/2 does not exist.
________________________________________________________________________________
web/gettext.ex:1:callback_type_mismatch
Callback mismatch for @callback MACRO-ngettext_noop/3 in Gettext.Backend behaviour.

Expected type:
binary()

Actual type:
{{:., [], [AAUStats.Gettext | :dngettext_noop, ...]}, [], [any(), ...]}
________________________________________________________________________________
web/views/table_view.ex:248:invalid_contract
Invalid type specification for function.

Function:
AAUStats.TableView.get_data_entries/1

Success typing:
@spec get_data_entries([{_, _}]) :: [any()]
________________________________________________________________________________
```

[...]

PR #261, opened 2018-10-23
Fix legacy warning filtering.

asummers:
Resolves https://github.com/jeremyjh/dialyxir/issues/260

This area of the code is much hackier than it has any right to be, but I think fixing in this PR would obfuscate the fix. We should follow up with a cleanup of some of these concepts.


jeremyjh:
Also one issue I believe is that it is evaluating the `exs` file for each warning. But I agree with just fixing this and then doing another PR to refactor.

PR #267, opened 2018-10-31
s/fun/function/

tiagopog:
Today I went through an error where I got the following message:

```
lib/ziad_ex/payload.ex:44:no_return
The created fun has no local return.
```

I've got the meaning of `fun` by the context, I think it's worth making it a bit clearer though.

BTW, thanks for this awesome lib :-)

### DIFF
--- a/lib/dialyxir/warnings/no_return.ex
+++ b/lib/dialyxir/warnings/no_return.ex
@@ -15,7 +15,7 @@ defmodule Dialyxir.Warnings.NoReturn do
     name_string =
       case name do
         [] ->
-          "The created fun"
+          "The created function"
 
         [function, arity] ->
           "Function #{function}/#{arity}"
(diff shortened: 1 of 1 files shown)


jeremyjh:
I think what we are after with `fun` is that it's an anonymous function, but this is Erlang terminology. Personally I think `fn` would be more clear, but maybe we should just say `anonymous function`. Thoughts @asummers ?

asummers:
In this case, it's exactly anonymous function. But I'm fine with either verbiage -- either yours or the one currently in PR. This message is derived from the original Erlang messaging and hasn't gotten an editor's hand yet, apparently.

jeremyjh:
@tiagopog do you mind updating to say `anonymous function` ?

jeremyjh:
Thanks!

Issue #248, opened 2018-09-18
No explanation for call_without_opaque warning


AndrewDryga:
Here is warning example:
```
apps/talkinto_domain/lib/talkinto_domain/transports/events.ex:52:call_without_opaque
The call Ecto.Multi.run(_multi@1::#{'__struct__':='Elixir.Ecto.Multi', _=>_},'event',fun((_,_) -> any())) does not have an opaque term of type 'Elixir.Ecto.Multi':t() in 1.
```

AndrewDryga:
Code example that causes this issue (with Ecto master):
```
  def insert_transport_delete_event(%Multi{} = multi, %Subject{user: %User{} = user}) do
    Multi.run(multi, :event, fn repo, %{transport: transport} ->
      %Event{
        resource: "TalkInto.Domain.Transports.Transport",
        transport_id: transport.id,
        account_id: transport.account_id || transport.account.id,
        action: "delete",
        actor_type: "user",
        actor_id: user.id
      }
      |> repo.insert()
    end)
  end
```

asummers:
#270 will resolve this, though I don't love the copy I wound up with. It can be iterated on.

asummers:
This is now merged into master.

PR #270, opened 2018-11-06
Call without opaqueness copy.

asummers:
I don't love these explanations at all, but this is way better than what is there. We can iterate. 

See: #248 

E.g.:

```
lib/demolyzer.ex:10:call_without_opaque
Function call without opaqueness type mismatch.

Call does not have expected opaque term of type OpaqueStruct.t() in the 1st position.

Example.do_error(_struct :: %OpaqueStruct{_ => _})
```

### DIFF
--- a/lib/dialyxir/warnings/call_without_opaque.ex
+++ b/lib/dialyxir/warnings/call_without_opaque.ex
@@ -8,44 +8,70 @@ defmodule Dialyxir.Warnings.CallWithoutOpaque do
   @impl Dialyxir.Warning
   @spec format_short([String.t()]) :: String.t()
   def format_short(_) do
-    "Call without opaqueness type mismatch."
+    "Function call without opaqueness type mismatch."
   end
 
   @impl Dialyxir.Warning
   @spec format_long([String.t()]) :: String.t()
   def format_long([module, function, args, expected_triples]) do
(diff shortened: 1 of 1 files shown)


mezz:
Hello, I am running into this warning and am not sure what it is telling me.
Can you please provide an example where it is bad, and an example where the issue is fixed?

jeremyjh:
There is an example in this commit, which you can also see by running `mix dialyzer.explain call_without_opaque`.

```elixir
  defmodule OpaqueStruct do
      defstruct [:opaque]

      @opaque t :: %OpaqueStruct{}
    end

    defmodule Example do
      @spec error(OpaqueStruct.t()) :: :error
      def error(struct = %OpaqueStruct{}) do
        do_error(struct)
      end

      @spec do_error(OpaqueStruct.t()) :: :error
      defp do_error(_) do
        :error
      end
    end
```

The purpose of opaque is to provide encapsulation violation warnings. The problem in the example is that we're directly constructing OpaqueStruct, instead we need to create and call a constructor function that is defined with a spec like `@spec make_opaque(params) :: OpaqueStruct.t`.

mezz:
Thanks for the info! That makes sense to me now.
I think I was getting this error due to some unrelated problem which confused me, I may open a new issue when I figure out exactly what I was doing wrong.

PR #272, opened 2018-11-08
Add ignore warnings default file behavior.

### LINKED ISSUE #224: Default ignore file?
Thinking of making `.dialyzer_ignore.exs` a default ignore file - if the file exists, we use it for ignores unless a different file is configured as the ignore file. This is something else people trip over. Also should probably update docs to make this more clearly the preferred format.

asummers:
Resolves #224 

Also took the opportunity to clean up some things in these files, particularly naming of functions to signify legacy where appropriate. Messages user when using the default.

### DIFF
--- a/lib/mix/tasks/dialyzer.ex
+++ b/lib/mix/tasks/dialyzer.ex
@@ -160,12 +160,28 @@ defmodule Mix.Tasks.Dialyzer do
           check_plt(force_check?)
         end
 
-      unless Dialyxir.Project.dialyzer_ignore_warnings() do
-        info("""
-        No :ignore_warnings opt specified in mix.exs. Using default: #{
-          Dialyxir.Project.default_ignore_warnings()
-        }
-        """)
+      default = Dialyxir.Project.default_ignore_warnings()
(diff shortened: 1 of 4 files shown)


lib/mix/tasks/dialyzer.ex
```diff
@@ -160,6 +160,14 @@ defmodule Mix.Tasks.Dialyzer do
           check_plt(force_check?)
         end
 
+      unless Dialyxir.Project.dialyzer_ignore_warnings() do
+        info("""
```
jeremyjh: I think we should only echo this if the default file exists - otherwise a user may misname the file (such as `dialyzer_ignore.exs` or `.dialyzer-ignore.exs`) think they've named it correctly and take this message as affirmation of that.  Also after your last comment I was actually thinking we should just always echo the ignore file in use, whether its the default or configuration.

asummers: Yeah I can do that. We could also reasonably check if the non default file exists here, too. I don't feel strongly any particular way so I'll make those changes.

asummers:
I updated the code with better messaging.

jeremyjh:
Looks great, thanks.

Issue #230, opened 2018-08-03
Invalid global installation instructions in README


bblaszkow06:
Hello!
Copy-pasting the current instructions for global installation result in an error because `deps.get` is missing before `compile` tasks.
It should probably look like this:
```bash
git clone https://github.com/asummers/erlex
cd erlex
mix do deps.get, compile, archive.build, archive.install
cd -
git clone https://github.com/jeremyjh/dialyxir
cd dialyxir
MIX_ENV=prod mix do deps.get compile, archive.build, archive.install
```
and for Windows:

```cmd
git clone https://github.com/jeremyjh/dialyxir
cd dialyxir
set "MIX_ENV=prod" && mix do deps.get, compile, archive.build, archive.install
```
Also, the installation may fail if older version has already been installed, so 
```
mix archive.uninstall dialyxir-**put_version_here**
```
might be necessary

Fl4m3Ph03n1x:
Created a PR to fix this issue:
https://github.com/jeremyjh/dialyxir/pull/275

PR #275, opened 2018-11-13
Fixing Global installation instructions in README

Fl4m3Ph03n1x:
Improving README installation instructions and per request of issue : 
 - https://github.com/jeremyjh/dialyxir/issues/230

Since it has been several months I don't think the original author plans on making a PR, so I did it myself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ If you are planning to use Dialyzer with an application built with the [Phoenix 
 
 ## Installation
 
+### Project installation
+
 Dialyxir is available on [hex.pm](https://hex.pm/packages/dialyxir).
 
 You can either add it as a dependency in your mix.exs, or install it globally as an archive task.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +22,16 @@ If you are planning to use Dialyzer with an application built with the [Phoenix
 
 ## Installation
 
+Before installing dialyxir make sure you have the correct version of elixir installed.
+You should have an elixir version with the otp bundle inside like `X-otp-Y`, where `X` is an elixir version, like 1.7.3 and Y an erlang version, like 21.
+For example, the following install command with `asdf` would install a compatible version with dialyxir:
+
+- `asdf install elixir 1.7.3-otp-21`
+
+However the command `asdf install elixir 1.7.3` would install an incompatible version, which would result in several warnings.
```
NobbZ: I'm not sure if we need instructions about how to install elixir in this document.

Especially as there are many things to consider. Operating system used? Version manager used? Something else…

I do agree though, that a warning should be added that mentions the problems when you have an elixir compiled with an older version of OTP vs running on a recent one.

But that again should be added to a troubleshooting section then and not under installation.

Fl4m3Ph03n1x: I believe asdf also works for windows? I might be wrong. If that's not the case I will need to change this. 

However, I do think that the **Installation** section is the perfect place for instructions on how to install elixir compatible versions with dialyxir. I added this in response to my issue and the issue Ben Smith was having ( https://elixirforum.com/t/dialyzer-not-compiling-correctly/17873/10?u=fl4m3ph03n1x )

Basically, I have the feeling that this is something that happens often (it also happened to someone else in my company), but people just feel discouraged to push forward and fix it and thus they just leave dialyzer behind. 

The problem is not how to install elixir _per se_, but which versions one should have. If this is not clear in my suggestion, then I totally I agree with you that I need to change it.

Fl4m3Ph03n1x: Given the favoritism to your other PR, and the lack of attention from the owners of the library to this PR, I have decided to close it in favor of your PR, which seems to have gotten general consensus.

Issue #274, opened 2018-11-13
New release candidate not working on sample project


Fl4m3Ph03n1x:
### Environment

* Elixir & Erlang/OTP versions (elixir --version):

output of `asdf current`
```
elixir         1.7.3-otp-21(set by /home/fl4m3/Workspace/elixir_and_otp_guidebook/dialyxyr_playground/.tool-versions)
erlang         21.1    (set by /home/fl4m3/.tool-versions)
```

* Which version of Dialyxir are you using? (cat mix.lock | grep dialyxir):

`[{:dialyxir, "~> 1.0.0-rc.4", only: [:dev], runtime: false}]`

### Current behavior

Steps taken:
1. `mix new dialyxir_playground_v2`
2. `cd dialyxir_playground_v2`
3. `asdf local elixir 1.7.3-otp-21`
4. add dependency `[{:dialyxir, "~> 1.0.0-rc.4", only: [:dev], runtime: false}]`
5. `mix do deps.get, deps.compile`

This is where I get the first issue. A ton of warnings, like with the 0.5 release, but more.

```
Resolving Hex dependencies...
Dependency resolution completed:
Unchanged:
  dialyxir 1.0.0-rc.4
  erlex 0.1.6
All dependencies up to date
==> erlex
Compiling 1 file (.yrl)
src/parser.yrl: Warning: conflicts: 23 shift/reduce, 0 reduce/reduce
Compiling 1 file (.xrl)
Compiling 2 files (.erl)
Compiling 1 file (.ex)
Generated erlex app
==> dialyxir
Compiling 56 files (.ex)
warning: redefining module Dialyxir.Warnings.FuncionApplicationArguments (current version loaded from /home/fl4m3/.asdf/installs/elixir/1.7.3-otp-21/.mix/archives/dialyxir-1.0.0-rc.4/dialyxir-1.0.0-rc.4/ebin/Elixir.Dialyxir.Warnings.FuncionApplicationArguments.beam)
  lib/dialyxir/warnings/function_application_arguments.ex:1

warning: redefining module Dialyxir.Warnings.UnknownType (current version loaded from /home/fl4m3/.asdf/installs/elixir/1.7.3-otp-21/.mix/archives/dialyxir-1.0.0-rc.4/dialyxir-1.0.0-rc.4/ebin/Elixir.Dialyxir.Warnings.UnknownType.beam)
  lib/dialyxir/warnings/unknown_type.ex:1

[...]

Issue #273, opened 2018-11-13
dialyxir not working for simple project


Fl4m3Ph03n1x:
### Environment

* Elixir & Erlang/OTP versions (elixir --version):

output of `asdf current`
```
elixir         1.7.3-otp-21(set by /home/fl4m3/Workspace/elixir_and_otp_guidebook/dialyxyr_playground/.tool-versions)
erlang         21.1    (set by /home/fl4m3/.tool-versions)
```

* Which version of Dialyxir are you using? (cat mix.lock | grep dialyxir):

`[{:dialyxir, "~> 0.5", only: [:dev], runtime: false}]`

### Current behavior

I have created a small project to play around with dialyxir. The experience so far has been puzzling at best.

Steps taken:
1. `mix new dialyxir_playground`
2. `cd dialyxir_playground`
3. `asdf local elixir 1.7.3-otp-21`
4. add dependency `[{:dialyxir, "~> 0.5", only: [:dev], runtime: false}]`
5. `mix do deps.get, deps.compile`

This is where I get the first issue. A ton of warnings. I am not sure if this is normal, but it sure doesn't give me a lot of confidence.

```
Resolving Hex dependencies...
Dependency resolution completed:
Unchanged:
  dialyxir 0.5.1
All dependencies up to date
==> dialyxir
Compiling 5 files (.ex)
warning: redefining module Dialyxir.Output (current version loaded from /home/fl4m3/.asdf/installs/elixir/1.7.3-otp-21/.mix/archives/dialyxir-1.0.0-rc.4/dialyxir-1.0.0-rc.4/ebin/Elixir.Dialyxir.Output.beam)
  lib/dialyxir/output.ex:1

warning: redefining module Dialyxir.Plt (current version loaded from /home/fl4m3/.asdf/installs/elixir/1.7.3-otp-21/.mix/archives/dialyxir-1.0.0-rc.4/dialyxir-1.0.0-rc.4/ebin/Elixir.Dialyxir.Plt.beam)
  lib/dialyxir/plt.ex:5

warning: redefining module Dialyxir.Dialyzer (current version loaded from /home/fl4m3/.asdf/installs/elixir/1.7.3-otp-21/.mix/archives/dialyxir-1.0.0-rc.4/dialyxir-1.0.0-rc.4/ebin/Elixir.Dialyxir.Dialyzer.beam)
  lib/dialyxir/dialyzer.ex:1

[...]

PR #276, opened 2018-11-13
Remove instructions for global installation

NobbZ:
Installing dialyxir globally can cause trouble when working with projects that depend on dialyxir for themself as #273 and #274 have shown.

Also there is this comment by @josevalim suggesting to not promote installing archives too much as well:

https://github.com/elixir-lang/elixir/issues/7577#issuecomment-383300978

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,24 +38,6 @@ end
 mix do deps.get, deps.compile
 ```
 
-To install globally as an archive:
-
-```console
-git clone https://github.com/asummers/erlex
-cd erlex
-mix do compile, archive.build, archive.install
-cd -
-git clone https://github.com/jeremyjh/dialyxir
-cd dialyxir
-MIX_ENV=prod mix do compile, archive.build, archive.install
-```
-or, in Windows:
-```console
-git clone https://github.com/jeremyjh/dialyxir
-cd dialyxir
-set "MIX_ENV=prod" && mix do compile, archive.build, archive.install
-```
-
 ## Usage
 
 Use dialyxir from the directory of the mix project you want to analyze; a PLT file will be created or updated if required and the project will be automatically compiled.


Fl4m3Ph03n1x:
Although I agree global installations should be avoided, I don't think there is a huge problem provided the caveats of such approach are made evident:
 - https://github.com/jeremyjh/dialyxir/pull/275

NobbZ:
As `dialyxir` has a dependency, global installation should be discouraged even more.

There is no mechanism built into `mix` to automatically update installed archives or its dependencies, neither is there something that would track which of those were installed manually or as a dependency. So it will always be a huge pain to keep those up to date.

I'm strictly against proposing global installation for dialyxir. Especially after the dependency got added for v1.0.

Fl4m3Ph03n1x:
Alright, you convinced me. But what about all the people that have installed dialyxir globally? 
Perhaps we should instead add a statement that global installations are no longer supported with a troubleshoot section to help them uninstall globally?

jeremyjh:
I think I agree. The original global instructions predate hex. I'm not sure that there are really any good reasons to do it anymore, and of course removing the instructions doesn't make it impossible.

asummers:
Yeah, I think this is a good idea. Archives have really only caused trouble for me. How do we want to handle expressing that this happened to people?

jeremyjh:
What do you have in mind, do you want to replace with a short disclaimer that we don't recommend global installs because issues?

asummers:
And also probably recommending to uninstall. But that seems more suitable for a changelog, I think.

NobbZ:
Even though I think you are right @Fl4m3Ph03n1x, I fear its not always as easy as `mix archive.uninstall dialyxir`.

You have been through this already. Something still left artifacts of dialyxir in your load path and it clashed.

Because of this I consider removing those artifacts individual tasks that everyone has to deal with on their own (or ask for help in the community).

Issue #277, opened 2018-11-19
Warn on unused ignore file entry


evnu:
The feature to ignore warnings can be very useful. Would it be possible to output a warning (maybe disabled by default, configurable in `mix.exs`) if an ignore pattern did not apply anymore? That would simplify keeping the ignore file nice and clean. I think this could for example be implemented by storing the patterns in a named ETS at the beginning, and removing patterns which were used successfully.

asummers:
I fully support a feature like this. Wanna try your hand at a PR for it? Might make sense to just pop it into the counter list that we're printing already, something like `Unnecessary skips: 3` perhaps? Or are you thinking it just prints the unnecessary warning itself?

asummers:
@jeremyjh do we want to support a `--unused-warnings-as-errors` type flag/config in mix.exs? Where if that hypothetical counter is nonzero, we emit a nonzero exit code.

jeremyjh:
Maybe `--unused_ignores-as-errors` ?

evnu:
I can give a PR a try. It will take me some days, though. Is the approach with ETS reasonable, or would you prefer a different approach?

jeremyjh:
I don't think ETS is necessary but I can see why it would be attractive.  Still, I think I'd approach it differently. First, you only need to support the `.exs` warnings.  If you have `filter_warning?` return the filter term (ignore) instead of true/false (Enum.find instead of Enum.any?), then `filter_warnings` can, instead of doing an `Enum.reject` it can `Enum.reduce` over a pair of structures: a map of used warnings, and a list of the passed warnings. 

Then you have just one more passed over the ignore file and print out any terms that aren't in the used map.

Another possibility - which would iron out another wart -  would be to start with a map of all the ignore file terms as keys and a value of 0 (since they haven't be used yet), and pass that all the way down to `filter_warning?` as its keys could be used directly instead of evaluating the file again for each warning. You still reduce over it and update the counts, and then just warn on any that were not used.

evnu:
> Might make sense to just pop it into the counter list that we're printing already, something like Unnecessary skips: 3 perhaps? Or are you thinking it just prints the unnecessary warning itself?

I think both is a good idea. Should this be configurable (e.g. `:list_unnecessary_filters` to enable listing them)?

Example output:

```
Total errors: 1, Skipped: 1, Unnecessary Skips: 1
Unused filters:
{"a/file.ex:17:no_return Function format_long/1 has no local return."}
```

PR #278, opened 2018-11-22
List unnecessary ignore filters

evnu:
This pull requests implements listing unnecessary ignore filters as proposed in #277.

### DIFF
--- a/lib/dialyxir/filter_map.ex
+++ b/lib/dialyxir/filter_map.ex
@@ -1,0 +1,71 @@
+defmodule Dialyxir.FilterMap do
+  @moduledoc """
+  A counters holding warnings to be skipped.
+
+  `:counters` points to a `Map` where the keys are warnings to be skipped and the value indicates
+  how often the warning was skipped.
+  """
+  defstruct list_unused_filters?: false, unused_filters_as_errors?: false, counters: %{}
+
+  @doc """
+  Fill a `FilterMap` from an ignore file.
+  """
+  def from_file(ignore_file, list_unused_filters?, halt_exit_status?) do
+    filter_map = %__MODULE__{
+      list_unused_filters?: list_unused_filters?,
+      unused_filters_as_errors?: list_unused_filters? && halt_exit_status?
+    }
+
+    {ignore, _} =
+      ignore_file
+      |> File.read!()
+      |> Code.eval_string()
+
+    Enum.reduce(ignore, filter_map, fn skip, filter_map ->
+      put_in(filter_map.counters[skip], 0)
+    end)
+  end
+
+  @doc """
+  Remove all non-allowed arguments from `args`.
+  """
+  def to_args(args) do
+    Keyword.take(args, [:list_unused_filters, :halt_exit_status])
+  end
+
+  @doc """
+  Retrieve the filters from a `FilterMap`.
+  """
+  def filters(filter_map) do
+    Map.keys(filter_map.counters)
+  end
+
+  @doc """
+  Increase usage count of a filter in `FilterMap`.
+  """
+  def inc(filter_map, filter) do
+    update_in(filter_map.counters[filter], &(&1 + 1))
+  end
+
+  @doc """
+  List unused filters.
+  """
+  def unused_filters(filter_map) do
+    filter_map.counters
+    |> Enum.filter(&unused?/1)
+    |> Enum.unzip()
+    |> elem(0)
+  end
+
+  @doc """
+  Determine if any filters were not used.
+  """
+  def unused_filters?(filter_map) do
+    Enum.any?(filter_map.counters, &unused?/1)
+  end
+
+  @doc """
+  Check if a `FilterMap` entry is unused.
+  """
+  def unused?({_filter, count}), do: count < 1
+end
--- a/lib/dialyxir/formatter.ex
+++ b/lib/dialyxir/formatter.ex
@@ -7,22 +7,26 @@ defmodule Dialyxir.Formatter do
   """
   import Dialyxir.Output, only: [info: 1]
 
+  alias Dialyxir.FilterMap
+
   def formatted_time(duration_us) do
     minutes = div(duration_us, 60_000_000)
     seconds = (rem(duration_us, 60_000_000) / 1_000_000) |> Float.round(2)
     "done in #{minutes}m#{seconds}s"
   end
 
-  def format_and_filter(warnings, _, :raw) do
+  def format_and_filter(warnings, _, _filter_map_args, :raw) do
     Enum.map(warnings, &inspect/1)
   end
 
-  def format_and_filter(warnings, filterer, :dialyxir) do
+  def format_and_filter(warnings, filterer, filter_map_args, :dialyxir) do
     divider = String.duplicate("_", 80)
+    filter_map = filterer.filter_map(filter_map_args)
+
+    {formatted_warnings, filter_map} = filter_warnings(warnings, filterer, filter_map)
 
     formatted_warnings =
-      warnings
-      |> filter_warnings(filterer)
+      formatted_warnings
       |> filter_legacy_warnings(filterer)
       |> Enum.map(fn warning ->
         message =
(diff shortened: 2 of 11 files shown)


evnu:
I pushed this early to get some more feedback. The current implementation allows enabling/disabling listing the filters with `:list_unused_filters`. It does not yet implement a "warnings as error" flag, as I was not sure were to implement that properly.

Example output with `:list_unused_filters` enabled:

```
Total errors: 1, Skipped: 1, Unnecessary Skips: 1
Unused filters:
{"a/file.ex:17:no_return Function format_long/1 has no local return."}
```

The pull request also misses an update to the README. I will do after the feature is complete.

lib/dialyxir/filter_map.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Dialyxir.FilterMap do
+  @moduledoc """
+  A map holding warnings to be skipped.
+
+  `:map` points to a `Map` where the keys are warnings to be skipped and the value indicates
+  how often the warning was skipped.
+  """
+  defstruct list_unused_filters?: false, map: %{}
```
jeremyjh: Maybe call the `map` member `counters` ?

jeremyjh:
This looks good so far.

lib/dialyxir/filter_map.ex
evnu: Done with 49795bb.

evnu:
71fad58 is a first attempt at implementing "warnings as errors". By setting `:unused_filters_as_errors`, dialyxir should return with an error code if unused filters are found. @jeremyjh What do you think, is this the right direction?

evnu:
> :unused_filters_as_errors

I think a flag like this should also imply that the unused filters are listed. Otherwise, the output is probably confusing.

lib/dialyxir/formatter.ex
```diff
@@ -152,15 +173,41 @@ defmodule Dialyxir.Formatter do
...
+    info(
+      "Total errors: #{warnings_count}, Skipped: #{skipped_count}, " <>
+        "Unnecessary Skips: #{unnessary_skips_count}"
+    )
 
     :ok
   end
 
+  defp list_unnecessary_skips(filter_map) do
+    if filter_map.list_unused_filters? do
```
asummers: I think if you matched this in the head, it would read more cleanly. Additionally, convert the if ! to an `unless`, and prefer `Enum.map_join/3` to `Enum.map |> Enum.join`

lib/dialyxir/filter_map.ex
```diff
@@ -0,0 +1,59 @@
...
+  def inc(filter_map, filter) do
+    update_in(filter_map.counters[filter], &(&1 + 1))
+  end
+
+  @doc """
+  List unused filters.
+  """
+  def unused_filters(filter_map) do
+    filter_map.counters
+    |> Enum.filter(fn {_filter, count} -> count < 1 end)
```
asummers: This filter is used in a few places. Maybe name the predicate?

lib/dialyxir/formatter.ex
```diff
@@ -152,15 +173,41 @@ defmodule Dialyxir.Formatter do
...
+          |> Enum.join("\n")
+
+        info("Unused filters:\n#{unused_filters}")
+      end
+    end
+  end
+
+  defp count_unnecessary_skips(filter_map) do
+    filter_map.counters
+    |> Enum.reject(fn {_k, v} -> v > 0 end)
```
asummers: The predicate I mentioned above could be used here too, I think

lib/dialyxir/formatter.ex
evnu: 5777dc9

lib/dialyxir/filter_map.ex
evnu: c234daf

lib/dialyxir/formatter.ex
evnu: c234daf

lib/dialyxir/formatter.ex
evnu: Clause error fixed with 21c59ed (squashed into 5777dc9).

asummers:
This is looking really good. Going to give it another look this evening, but I didn't see anything that stuck out to me with your most recent changes.

asummers:
Seems like Dialyzer isn't honoring the status code, according to Travis.

lib/dialyxir/project.ex
```diff
@@ -77,23 +79,27 @@ defmodule Dialyxir.Project do
 
   defp skip?(_, _), do: false
 
-  def filter_warning?({file, warning, line, short_description}) do
+  def filter_warning?({file, warning, line, short_description}, filter_map = %FilterMap{}) do
+    {matching_filters, _non_matching_filters} =
+      Enum.split_with(
```
asummers: This might read nicely as a pipe.

evnu: d751787 ?

asummers: Perfect.

evnu:
>> Seems like Dialyzer isn't honoring the status code, according to Travis.
> I will check that.

4216113 changes this behaviour. By default, the status code is not changed even if unused filters are present. A user can make this an error with `unused_filters_as_errors`. Is this the behaviour we want here? With that, additionally listing the unnecessary filters might not be needed. Users will encounter output such as this:

```
Total errors: 26, Skipped: 26, Unnecessary Skips: 1
```

They will then wonder what `Unnecessary Skips: 1` means, referring to the README section (TBD) and reading about the two possible options `:list_unused_filters` and `:unused_filters_as_errors`, allowing them to choose how they want to handle this.

evnu:
af8f9e4 extends the README.

jeremyjh:
I'm struggling to think of an occasion where I'd want a non-zero status for unused filters yet not want to see which filters were unused. In other words I think `unused_filters_as_errors` should imply `list_unused_filters`.

evnu:
With 909904a, `unused_filters_as_errors` implies `list_unused_filters`. This is also tested in `formatter_test.exs`.

evnu:
Do you want the commits to be squashed and rebased? If so, should I do that before the review is finished, or afterwards?

jeremyjh:
@evnu yes please squash once we approve. you can rebase on master whenever you like. are there any tasks remaining on this PR?

evnu:
I rebased the branch and added command line arguments with c6da037. I think with this the TODOs w.r.t. functionality should be finished now.

jeremyjh:
Woops, one issue other than the formatting, is that it looks like the CLI arg is getting passed down to dialyzer which complains of an unknown warning:

https://travis-ci.org/jeremyjh/dialyxir/builds/464953658#L864

CLI args need to be captured in the [command options](https://github.com/jeremyjh/dialyxir/blob/master/lib/mix/tasks/dialyzer.ex#L132) so that they are split into a separate list from the dialyzer args (all unknown CLI args are passed to dialyzer with the assumption that they are specific warning flags).

evnu:
> LGTM, thanks so much for all your efforts on this!

My pleasure.

> Woops, one issue other than the formatting, is that it looks like the CLI arg is getting passed down to dialyzer which complains of an unknown warning:

Yes, another typo. Should be fixed now.

asummers:
@jeremyjh I wonder if we should warn on unknown opts. I’ll make an issue and we can discuss there.

evnu:
Squashed into f3c31c3.

jeremyjh:
It looks like its not halting with a non-zero exit code when we use --unused-filters-as-errors: 

The WX filter in our own ignore file is apparently unused, and while it prints the warning in red, it still exits with a 0:

https://travis-ci.org/jeremyjh/dialyxir/builds/464988166#L879

evnu:
> It looks like its not halting with a non-zero exit code when we use --unused-filters-as-errors

It does abort if `--halt-exit-status` is also used. I will try to figure out why that is currently also needed.

evnu:
>  I will try to figure out why that is currently also needed.

The exit status of `Dialyzer.dialyze/3` is ignored if `halt_exit_status` is not set.

jeremyjh:
Ah yes, that makes sense. Maybe we should note that in the doc for `unused_filters_as_errors` though?

evnu:
>  Maybe we should note that in the doc for unused_filters_as_errors though?

That would work, but I wonder if the option itself makes sense at all if `--halt-exit-status` is needed anyways. I suggest we drop the option and make `list_unused_filters` always result in a warning. That way, `--halt-exit-status` can be used directly together with that option. ea1cc70 implements just that and should fail due to the unused WX filter. What do you think, does this make sene?

evnu:
Adding to the previous comment, if `unused_filters_as_errors` is dropped, then `--halt-exit-status` should also imply `list_unused_filters`.

evnu:
With 33acdef, `halt-exit-status` now implies `:list_unused_filters`.

jeremyjh:
If `--halt-exit-status` without other options means unused filters will result in a non-zero exit then we'll break people's CI builds when they upgrade. Maybe they should be happy to fix them but I'm not sure its worth it, or that there should be no way to opt-out of unused warnings as errors in your CI process (which has to use --halt-exit-status). What *does* make sense to me is that `--list-unused-errors + --halt-exit-status` implies warnings as errors.

Also I noticed the warning isn't highlighted in red anymore: https://travis-ci.org/jeremyjh/dialyxir/builds/465303756#L858

evnu:
> What does make sense to me is that --list-unused-errors + --halt-exit-status implies warnings as errors.

Yes, that sounds reasonable. I think it would make sense to derive `:unused_filters_as_errors = --list-unused-errors && --halt-exit-status`, but not expose this as a user-definable variable. So I would go back to the original behaviour from https://github.com/jeremyjh/dialyxir/commit/f3c31c3050b54be1110f881c2a9bff734fa600b6 the squashed commit, remove documentation for `:unused_filters_as_errors ` and derive that value from the other parameters. Does that sound like the right approach?

> Also I noticed the warning isn't highlighted in red anymore:

Ah, yes, I think that is because it is `:warn`, not `:error`. This should be resolved as well with what I wrote further above (going back to the original behaviour etc).

jeremyjh:
> So I would go back to the original behaviour from f3c31c3 the squashed commit, remove documentation for :unused_filters_as_errors and derive that value from the other parameters. Does that sound like the right approach?

It does to me. @asummers are you good with this as well?

asummers:
Sounds fine.

evnu:
38ea9b4 derives error from `list-unused-filters + halt-exit-status`. With 612a373, skipped filters are now printed in red if that is an error. For 612a373, I needed to pull printing from `formatter.ex` into `dialyzer.ex`. I am not sure if that solution is pretty, but I did not want to pass color-coding into formatter.

jeremyjh:
lgtm

jeremyjh:
Thanks so much!

PR #294, opened 2019-02-13
Better explanation for `invalid_contract`

axelson:
"Invalid type specification" did not make sense to me.

### DIFF
--- a/lib/dialyxir/warnings/invalid_contract.ex
+++ b/lib/dialyxir/warnings/invalid_contract.ex
@@ -9,7 +9,7 @@ defmodule Dialyxir.Warnings.InvalidContract do
   @spec format_short([String.t()]) :: String.t()
   def format_short([module, function, arity, _signature]) do
     pretty_module = Erlex.pretty_print(module)
-    "Invalid type specification for function #{pretty_module}.#{function}/#{arity}."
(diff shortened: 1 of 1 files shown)


lib/dialyxir/warnings/invalid_contract.ex
```diff
@@ -39,11 +39,17 @@ defmodule Dialyxir.Warnings.InvalidContract do
...
+      def process(:ok) do
         :ok
       end
     end
+
+    This does not match because the success typing (what actually happens) is
+    that the function returns `:ok`, if and only if the function receives `:ok`,
+    but the @spec of the function says that it receives `:error` and returns
+    `:ok`, but actually when the function receives `:error` it throws a
+    `FunctionClauseError`
```
asummers: I think we can wordsmith this a bit. Maybe something like:

```
The @spec in this case claims that the function accepts a parameter :error but the 
function head only accepts :ok, resulting in the mismatch. 
```

axelson: Haha, that's much better!

asummers:
The format check is failing because the short error is > 80 chars. We can either shorten the message, or do a `"first" <> "second"` type thing.

axelson:
Updated to fix the formatting

axelson:
Great, thanks!

Issue #221, opened 2018-07-13
Protocols for built-in data types


RobertDober:
Hello, as requested by @josevalim I would like to report this issue ~https://github.com/eslint/eslint-scope/issues/39~ [corrected below] here.

It has been discussed here https://elixirforum.com/t/dialyzer-listed-not-implemented-protocols-as-unknown-functions/2099.

And here is the project demonstrating the point https://github.com/im-tollu/unhappy_dialyzer.

skwerlman:
seems like there's a copy-paste error in the above post; issue is here: elixir-lang/elixir#7708

jeremyjh:
Thanks for putting all these details together! I agree this needs to be addressed in `dialyxir`.

RobertDober:
@skwerlman thanx for fixing my error

asummers:
Seems like your team had fun with that `eslint` issue yesterday too 😄 😱

RobertDober:
This is strange the error seemed to be fixed in 1.7.1 and reappears in 1.7.2, are you aware of this?
Using 0.5.1

gpad:
@jeremyjh I have the same bug Elixir 1.6 and dialyzer-rc.4

asummers:
@gpad Does this issue occur if you run `mix compile --force` beforehand? Trying to see how it could have been fixed for @RobertDober even though we didn't really explicitly fix things.

asummers:
That's a decent enough workaround for now. But we should fix this here eventually, or provide a flag to do that for you, perhaps? At the very least reference in the README. /cc @jeremyjh

gpad:
@asummers sorry What I mean is that also removing the `_build` the error is still here ...

jeremyjh:
The issue is always going to be there because we do not include the consolidation path in the beam paths. Unless 1.7.1 had a bug where it was putting the consolidated files in the ebin folder I don't see how that could have been "fixed". As Jose noted though, we can't just add that path as then we'll have duplicates we - need to do some kind of merging or diffing of these folders.

jeremyjh:
Dialyxir RC has support Regex ignores. I don't think there is ever a reason to report on a __impl__ function does not exist error. In case you agree, you could solve this forever with something like:

```elixir
# .dialyzer_ignore.exs
[
  ~r/unknown function.*__impl__.*/
]
```

gpad:
[...]

PR #299, opened 2019-02-22
Use .beam files after protocol consolidation

bblaszkow06:
Hello! This PR is my proposal for solving #221.
Based on info from here https://github.com/elixir-lang/elixir/issues/7708 I've changed the args for dialyzer and instead of passing directories, I'm passing files list. The list is created in such a way that if files with the same name are present inside `dialyzer_paths()`, only the last one will be passed to the dialyzer. Combine that with consolidation path added by default after the compilation path and the issue seems solved.

A few notes, though:
* I haven't tested it with umbrella projects
* args list for dialyzer becomes enormous, I've added a limit of 10 for args inspection, but maybe some other approach should be taken (e.g. not printing the files at all or only with some flag)

### DIFF
--- a/lib/dialyxir/dialyzer.ex
+++ b/lib/dialyxir/dialyzer.ex
@@ -35,7 +35,7 @@ defmodule Dialyxir.Dialyzer do
         info("Starting Dialyzer")
 
         args
-        |> inspect(label: "dialyzer args", pretty: true)
+        |> inspect(label: "dialyzer args", pretty: true, limit: 10)
         |> info
 
         {duration_ms, result} = :timer.tc(&:dialyzer.run/1, [args])
--- a/lib/dialyxir/dialyzer.ex
+++ b/lib/dialyxir/dialyzer.ex
@@ -35,7 +35,7 @@ defmodule Dialyxir.Dialyzer do
         info("Starting Dialyzer")
 
         args
-        |> inspect(label: "dialyzer args", pretty: true, limit: 10)
+        |> inspect(label: "dialyzer args", pretty: true)
         |> info
 
         {duration_ms, result} = :timer.tc(&:dialyzer.run/1, [args])
--- a/lib/dialyxir/dialyzer.ex
+++ b/lib/dialyxir/dialyzer.ex
@@ -35,7 +35,7 @@ defmodule Dialyxir.Dialyzer do
         info("Starting Dialyzer")
 
         args
-        |> inspect(label: "dialyzer args", pretty: true)
+        |> inspect(label: "dialyzer args", pretty: true, limit: 8)
         |> info
 
         {duration_ms, result} = :timer.tc(&:dialyzer.run/1, [args])
--- a/lib/dialyxir/project.ex
+++ b/lib/dialyxir/project.ex
@@ -61,7 +61,7 @@ defmodule Dialyxir.Project do
     |> Enum.map(fn {_file, path} -> path |> to_charlist() end)
   end
 
-  def dialyzer_paths do
+  defp dialyzer_paths do
     paths = dialyzer_config()[:paths] || default_paths()
     excluded_paths = dialyzer_config()[:excluded_paths] || []
     Enum.map(paths -- excluded_paths, &String.to_charlist/1)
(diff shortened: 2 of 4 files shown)


lib/dialyxir/project.ex
```diff
@@ -217,7 +226,7 @@ defmodule Dialyxir.Project do
 
   defp default_paths() do
     reduce_umbrella_children([], fn paths ->
-      [Mix.Project.compile_path() | paths]
+      [Mix.Project.compile_path() | [Mix.Project.consolidation_path() | paths]]
```
asummers: This can be done as:

```
[Mix.Project.compile_path(), Mix.Project.consolidation_path() | paths]
```

lib/dialyxir/project.ex
```diff
@@ -52,6 +52,15 @@ defmodule Dialyxir.Project do
...
+  def dialyzer_files do
+    dialyzer_paths()
+    |> Enum.map(fn path ->
+      path |> Path.join("*.beam") |> Path.wildcard() |> Enum.map(&{Path.basename(&1), &1})
+    end)
+    |> Enum.reduce(%{}, fn entries, acc -> Map.merge(acc, Map.new(entries)) end)
+    |> Enum.map(fn {_file, path} -> path |> to_charlist() end)
+  end
+
   def dialyzer_paths do
```
asummers: Is this called outside of this file anymore? Can probably be made private if not.

asummers:
Dialyzer is failing pretty spectacularly, and I'm uncomfortable merging this until it's tested on umbrellas.

lib/dialyxir/dialyzer.ex
```diff
@@ -35,7 +35,7 @@ defmodule Dialyxir.Dialyzer do
         info("Starting Dialyzer")
 
         args
-        |> inspect(label: "dialyzer args", pretty: true)
+        |> inspect(label: "dialyzer args", pretty: true, limit: 10)
```
asummers: I wonder if this should be an opt of some sort instead

bblaszkow06: I believe there could be a `--debug`/`--verbose` that would print the files (`limit: :infinity`) and by default, the files wouldn't be printed at all. Anyway, I've reverted that change, I'll leave the decision up to you

bblaszkow06:
@asummers I've tested an umbrella project and it turned out that consolidation path can be included just once (as it's shared for the whole umbrella project).

The 'spectacular failure' is actually not a bug, it's a feature as those are valid `:underspec` warnings :stuck_out_tongue: 
It seems that this has been fixed in Elixir 1.8 (here I believe: https://github.com/elixir-lang/elixir/pull/7832).

I've added Elixir 1.8.1 to the build matrix on Travis (and fixed the exclusion rule by the way) and as you can see (https://travis-ci.org/jeremyjh/dialyxir/builds/498428066), the build succeeds for this version. I'm not quite sure what to do with the failing versions - should they be removed? Or 'fixed' by silencing the warnings in `.dialyzer_ignore.exs`?

lib/dialyxir/dialyzer.ex
asummers: Exactly my thought. Adding that opt is probably out of scope of this anyway, but good call on bringing that up.

jeremyjh:
I think we're going to have to do something about the output including the entire files collection, this will be many pages of useless spam in larger projects.

Also, we shouldn't be getting warnings from Elixir lib files in our results should we? 
For example one of the warnings is in this file: `/home/build/elixir/lib/elixir/lib/list/chars.ex:1:contract_supertype`

I think its because this file is included: 
   '/home/travis/build/jeremyjh/dialyxir/_build/prod/lib/dialyxir/consolidated/Elixir.String.Chars.beam',

jeremyjh:
@bblaszkow06 are you planning to finish this soon? I'd like to get into 1.0 and to me it seems like it is pretty close. Let me know if you need help.

bblaszkow06:
@jeremyjh Sorry, I haven't had much time to finish this, but I managed to find a way to prevent seeing errors from Elixir lib files by taking from consolidation path only the files that were present in compilation path(s).
Now I have to clean up the code I was experimenting on and I hope I will find time for that by the end of this week.

bblaszkow06:
@jeremyjh All green and ready for review. :smile: There are some warnings printed for `MIX_ENV=prod mix dialyzer --format raw` test, but they are present on master as well

jeremyjh:
Looks good to me, thank you!

PR #296, opened 2019-02-15
Add plt_ignore_apps option to ignore dependencies

the-mikedavis:
Closes #295.

The documentation will need a small change as well. I'll leave that to you

PR #300, opened 2019-02-27
Ensure plt_ignore_apps option works in umbrella

gilacost:
[This PR](https://github.com/jeremyjh/dialyxir/pull/296) does not work for umbrella projects because the plt_ignored_apps are subtracted before calling `Enum.uniq()` in plt list.

A fixture reproducing this situation has been added to ensure that this works.


jeremyjh:
Thanks! I'd prefer to keep the fixtures really lean though, could you strip it down to the bare essentials for this test? Hopefully that is just three `mix.exs` files.

gilacost:
> Thanks! I'd prefer to keep the fixtures really lean though, could you strip it down to the bare essentials for this test? Hopefully that is just three `mix.exs` files.

Is it ok like this?

jeremyjh:
@gilacost Thanks again!

PR #320, opened 2019-04-19
Halt on warnings by default, offer --ignore-exit-status to override.

### LINKED ISSUE #315: Feedback Request: Make --halt-exit-status the default?
I wish I'd opened this earlier, but my intention is to try not to have any breaking changes after 1.0, and lately I've been thinking we really have the wrong default around `--halt-exit-status` and this is our last chance to address it. With good dialyzer support in ElixirLS that can perform incremental analysis and provide results in the editor, I think one of the most important use cases remaining for `dialyxir` is to run analysis in CI and we have the wrong default for that.

Command line usage is important too, but there is no harm in exiting with non-zero there that I can think of. The only reason we didn't make it the default to begin with is I thought some people might be using dialyxir in mix-watch or similar, and didn't want to break them by terminating the VM.  Still, that would be a pretty obvious issue,  that would happen after upgrading dialyxir, and we can support `--watch` CLI arg and config entries to enable those use-cases.

I'd prefer new users not setup a CI pipeline and only find out it isn't working weeks later when lots of errors have built up, which could easily happen since you'd have a clean run to start with, you may not realize a failure isn't failing CI.

I think we will not warn if the `--halt-exit-status` *is* provided unnecessarily; one feature we'd support is setting `watch` in your mix.exs and then overriding with the CLI arg.

Please respond with your +/- or comment and let me know of any other considerations I'm missing.

jeremyjh:
resolves #315 

Implements the suggestions from #315. Halting with a non-zero RC is now the default behavior with an `--ignore-exit-status` option that prevents a halt. When halting, an extra informational message is written. There is no halt if there are no warnings.

image

Also, there is a warning if the old `--halt-exit-status` option is provided. 

Right now it looks like this:

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ script:
   - MIX_ENV=prod mix dialyzer --list-unused-filters
   - OUTPUT_TESTS=true mix test
   - MIX_ENV=prod mix dialyzer --format short
-  - MIX_ENV=prod mix dialyzer --format raw
+  - MIX_ENV=prod mix dialyzer --format raw --ignore-exit-status
   - MIX_ENV=prod mix dialyzer --format dialyzer
 branches:
   except:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ script:
   - mix test --trace
   - MIX_ENV=prod mix do compile --warnings-as-errors, archive.build, archive.install --force
   - mix format --check-formatted
-  - MIX_ENV=prod mix dialyzer --list-unused-filters --halt-exit-status
+  - MIX_ENV=prod mix dialyzer --list-unused-filters
   - OUTPUT_TESTS=true mix test
   - MIX_ENV=prod mix dialyzer --format short
   - MIX_ENV=prod mix dialyzer --format raw
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: elixir
 elixir:
-  - 1.6.3
   - 1.7.3
   - 1.8.1
 otp_release:
@@ -8,7 +7,6 @@ otp_release:
   - 21.3
 matrix:
   exclude:
-    - elixir: 1.6.3
       otp_release: 21.3
 script:
   - mix test --trace
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ script:
   - MIX_ENV=prod mix dialyzer --list-unused-filters
   - OUTPUT_TESTS=true mix test
   - MIX_ENV=prod mix dialyzer --format short
-  - MIX_ENV=prod mix dialyzer --format raw --ignore-exit-status
+  - MIX_ENV=prod mix dialyzer --format raw
   - MIX_ENV=prod mix dialyzer --format dialyzer
 branches:
   except:
--- a/lib/dialyxir/dialyzer.ex
+++ b/lib/dialyxir/dialyzer.ex
@@ -6,12 +6,12 @@ defmodule Dialyxir.Dialyzer do
   alias Dialyxir.FilterMap
 
   defmodule Runner do
-    @dialyxir_args ~w(
-      raw
-      format
-      list_unused_filters
-      ignore_exit_status
-    )a
+    @dialyxir_args [
+      :raw,
+      :format,
+      :list_unused_filters,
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -14,10 +14,10 @@ script:
   - mix test --trace
   - MIX_ENV=prod mix do compile --warnings-as-errors, archive.build, archive.install --force
   - mix format --check-formatted
-  - MIX_ENV=prod mix dialyzer --list-unused-filters --halt-exit-status
+  - MIX_ENV=prod mix dialyzer --list-unused-filters
   - OUTPUT_TESTS=true mix test
   - MIX_ENV=prod mix dialyzer --format short
-  - MIX_ENV=prod mix dialyzer --format raw
+  - MIX_ENV=prod mix dialyzer --format raw --ignore-exit-status
```
jeremyjh: Someone noted this in a recent PR, but warnings are always printed in format raw. I don't know if that was intended,   or is just a consequence of not running it through the formatter? I'm not sure its a real problem but could be a bit confusing.

asummers: Wait, I'm not positive what you mean. Do you have a link to that discussion?

lib/dialyxir/dialyzer.ex
```diff
@@ -6,10 +6,16 @@ defmodule Dialyxir.Dialyzer do
   alias Dialyxir.FilterMap
 
   defmodule Runner do
+    @dialyxir_args ~w(
```
asummers: I would drop the sigil here, personally, but your discretion =)

lib/mix/tasks/dialyzer.ex
```diff
@@ -135,18 +135,23 @@ defmodule Mix.Tasks.Dialyzer do
...
-    plt: :boolean,
-    quiet: :boolean,
-    raw: :boolean,
-    format: :string
+
+  @old_options [
+    halt_exit_status: :boolean
   ]
 
+  @command_options Keyword.merge(@old_options,
```
asummers: This list is kind of long now, I would ABC it.

lib/mix/tasks/dialyzer.ex
```diff
@@ -257,14 +263,18 @@ defmodule Mix.Tasks.Dialyzer do
...
 
     {status, exit_status, [time | result]} = Dialyzer.dialyze(args)
     info(time)
     report = if status == :ok, do: &info/1, else: &error/1
     Enum.each(result, report)
-    if opts[:halt_exit_status], do: :erlang.halt(exit_status)
+
+    unless exit_status == 0 || opts[:ignore_exit_status] do
+      error("Halting VM with exit status #{exit_status}")
+      :erlang.halt(exit_status)
```
asummers: Why not https://hexdocs.pm/elixir/System.html#halt/1

.travis.yml
jeremyjh: It wasn't really discussed, it was noted here: https://github.com/jeremyjh/dialyxir/pull/299#issuecomment-483339798

The issue is just that with raw, all warnings are always printing and now it will exit with a non-zero, even if every warning is ignored. You can see it here: https://travis-ci.org/jeremyjh/dialyxir/jobs/522309081#L663

lib/dialyxir/dialyzer.ex
jeremyjh: I guess we should be consistent with it. The thing I like about it, is that diffs are cleaner as there is no issue with trailing commas. But its kind of an anachronism from Ruby probably.

lib/mix/tasks/dialyzer.ex
jeremyjh: Sorry ABC?

asummers: Alphabetize =)

asummers: Makes finding things as a reader slightly simpler, and makes it obvious where to insert new ones.

.travis.yml
asummers: Oh that's funky. That seems like a bug.

asummers: Looking into fixing.

jeremyjh:
@leepage87 published 1.0.0-rc.7

leepage87:
Thanks!

PR #322, opened 2019-04-21
Improve verbiage in several warnings.

asummers:
This is the first of probably 3 PRs to fix some of the verbiage in the short messages to be more useful.

### DIFF
--- a/lib/dialyxir/warnings/call.ex
+++ b/lib/dialyxir/warnings/call.ex
@@ -7,8 +7,10 @@ defmodule Dialyxir.Warnings.Call do
 
   @impl Dialyxir.Warning
   @spec format_short([String.t()]) :: String.t()
-  def format_short([_module, function | _]) do
-    "The function call #{function} will not succeed."
+  def format_short([module, function | _]) do
+    pretty_module = Erlex.pretty_print(module)
+
+    "The function call #{pretty_module}.#{function} will not succeed."
   end
 
   @impl Dialyxir.Warning
--- a/lib/dialyxir/warnings/call_to_missing_function.ex
+++ b/lib/dialyxir/warnings/call_to_missing_function.ex
@@ -7,13 +7,15 @@ defmodule Dialyxir.Warnings.CallToMissingFunction do
 
   @impl Dialyxir.Warning
   @spec format_short([String.t()]) :: String.t()
-  def format_short(args), do: format_long(args)
+  def format_short([module, function, arity]) do
+    pretty_module = Erlex.pretty_print(module)
+    "Call to missing or private function #{pretty_module}.#{function}/#{arity}."
+  end
 
   @impl Dialyxir.Warning
   @spec format_long([String.t()]) :: String.t()
-  def format_long([module, function, arity]) do
-    pretty_module = Erlex.pretty_print(module)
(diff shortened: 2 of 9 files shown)


jeremyjh:
The reason I did not include `module` in most of these warnings when I updated them, is:

1. The file name is already part of the warning printed on the screen. 
2. The main use case for `short` is in IDEs/editors, where you are already in that file.
3. Even medium length module paths can easily push us well over 100 characters.

jeremyjh:
@asummers any comment on this? do you still plan to submit more verbiage changes?

asummers:
Hey, sorry, I've been MIA the past few weeks. Locally, I prefer to look at the short dialyzer warnings, even in CLI. I guess you're right that in editors that's less helpful. Should we consider adding an intermediary length format (which can be pushed past V1) between short and long (medium?) that does short + module names? Maybe that's only useful to me; unsure. I just like being able to quickly scan a list  to see what's broken, without getting the entire full output with divider lines and such.

jeremyjh:
I guess I'm not opposed to an intermediate length warning. I'm not opposed to all of these changes either, I just would like to agree we are looking at "suitability" for inclusion of info in a short warning through the lens of:

1. Enough info to disambiguate the warning in ignore files without depending on a line number.
2. Enough info to identify the culprit or error subject from the context provided by the (source code at the) call site and the message in the error.
3. Trying to squeeze this into 100 chars in typical code for display in editor status lines (which will fail no matter what in some cases).

Where 1 & 2 are satisfied with less info but you'd like to see that info at the CLI, the info should be included if we are likely good on 3.

Inclusion of module/callback names tends to lead us into trouble on point 3 because Elixir module names are so long. Maybe we need to use a short name that is just the last part of the module name. For example if you run `MIX_ENV=examples mix format --short` in this branch you'll see the warning for callback_arg_type_mismatch is now 172 characters. Its not terribly short to begin with but you can see the behavior name really stretches it:


> test/examples/callback_argument.ex:8:callback_arg_type_mismatch Type mismatch with behaviour callback Dialyxir.Examples.CallbackArgument.Behaviour.ok/1 in the 1st argument

asummers:
So I guess do we want to go to war with this then? And then add the new error in a followup? Do we want to change the name `short` (I don't think we do, but posing it).

jeremyjh:
If you wanted to use a `short_module` function for module and behaviour names included in the short warnings I'd be ok with that, and then maybe we wouldn't need another format? But if what you really want is the freedom to output 2-3 lines then yeah we can ship whats in master and add another format later.

jeremyjh:
@asummers I actually think goes a long way towards fixing it. If we [unqualify](https://github.com/jeremyjh/dialyxir/commit/74dd4a045cac4880f6493c4384dae9d8c3e0cca3) the module/behaviour names in the short warnings they are quite a lot shorter.

The callback example is still 137 chars, but we're taking a lot of characters from the warning name itself there, which I don't see any option to improve.

`test/examples/callback_argument.ex:8:callback_arg_type_mismatch Type mismatch with behaviour callback Behaviour.ok/1 in the 1st argument.`

jeremyjh:
> So I guess do we want to go to war with what's in master then? And then add the new error in a followup? Do we want to change the name short (I don't think we do, but posing it).

Yes, closing this.

PR #334, opened 2019-06-03
Remove pattern print as it overflows with long patterns.

asummers:
On sufficiently long patterns, this overflows lines.


jeremyjh:
If there is no pattern information it isn't possible to ignore this warning without referencing a specific line number. How bad is the overflow, do you have example output?

PR #347, opened 2019-07-08
Refactor coloring

sirikid:


test/dialyxir/dialyzer_test.exs
```diff
@@ -28,13 +30,8 @@ defmodule Dialyxir.DialyzerTest do
...
       assert expected_exit_code == 2
 
-      assert expected_messages == [
-               "time",
-               "\e[31mwarning 1\e[0m",
-               "\e[31mwarning 2\e[0m",
-               "",
-               "\e[33mdone (warnings were emitted)\e[0m"
-             ]
+      assert IO.chardata_to_string(expected_messages) ==
```
jeremyjh: I really don't find this to be as readable here. I think I'd prefer to leave the asserts the way they were.

lib/dialyxir/output.ex
```diff
@@ -2,20 +2,8 @@ defmodule Dialyxir.Output do
...
-          ANSI.yellow() <> text <> ANSI.reset()
-
-        :green ->
-          ANSI.green() <> text <> ANSI.reset()
-
-        _ ->
-          text
-      end
+    if color in [:red, :yellow, :green] do
+      ANSI.format([apply(ANSI, color, []), text, ANSI.reset()])
```
jeremyjh: Per the docs, a reset is included automatically.

https://hexdocs.pm/elixir/IO.ANSI.html#format/2

asummers: I don't find this to be better. If I add `:purple` to that list, we don't get compiler help, but in the current implementation `ANSI.purple()` would give compiler errors. I think explicit is better here, even if it is a bit more verbose.

sirikid:
image

sirikid:
There is no isomorphism between strings and chardatas, so I think this approach is error prone.

jeremyjh:
Thanks for your interest but I don't really think its an improvement, or at least not a meaningful one.

sirikid:
It is not an improvement, it is a bug fix.
repo: https://github.com/sirikid/dialyxir_bug
image

jeremyjh:
I wish you would mention that in the issue description, or at least the title. You just labelled it as refactoring.

jeremyjh:
Can you add a failing test? AFAIK we should be able to change `:ansi_disabled` flag at runtime.

sirikid:
Do we really need a test for stdlib?

jeremyjh:
If it were only up to the stdlib there would be no need for a bug-fix. Typically when we add a fix we first add a test that fails before the fix is applied. In this case that would mean you set `:ansi_disabled` yet the text still has ANSI codes.

sirikid:
Sorry, I am really busy now, will write the tests as soon as possible

sirikid:
It took longer than I expected

PR #374, opened 2019-11-15
Fix one format mention in README

kenny-evitt:
The `dialyzer` format doesn't include warning names but `raw` does.


kenny-evitt:
My bad – I totally missed the warning name in the default format output!

Issue #264, opened 2018-10-26
Ignore all files in a path


hylkealons:
I would like to disable dialyzer for all files in a folder. Is that currently possible?

jeremyjh:
If you use master, there is support for regular expressions in your ignore file. Take a look our own ignore file for an example.

The text it matches on, is the warning emitted with `mix dialyzer --format short`, so you can check that then write a regex. Make sure you add the ignore file to your config (README explains it pretty well).

 Probably it would be something like:

.dialyzer-ignore.exs
```elixir
[ 
~r/^/path/to/my_folder.*/
]

asummers:
Going to go ahead and close this since this is possible on master. If this is still an issue due to bugs, we can reopen.

PR #382, opened 2020-02-27
Add exclude files config option

juffel:
This adds a configurable `:exclude_file_suffixes` parameter that can be used to exclude certain .beam files.
It also updates the README and adds a reference to both the existing `:exclude_paths` and the new `:exclude_file_suffixes` param.

Not so sure about testing this, since it is tightly coupled to `dializer_files/0` which directly reads files from the file system 🤔

### DIFF
--- a/lib/dialyxir/project.ex
+++ b/lib/dialyxir/project.ex
@@ -77,7 +77,7 @@ defmodule Dialyxir.Project do
 
     Enum.reject(files, fn file ->
       Enum.any?(file_exclusions, fn exclusion ->
-        String.match?(file, ~r/#{exclusion}$/)
+        String.ends_with?(file, exclusion)
       end)
     end)
   end
--- a/lib/dialyxir/project.ex
+++ b/lib/dialyxir/project.ex
@@ -67,7 +67,19 @@ defmodule Dialyxir.Project do
 
     beam_files
     |> Map.merge(consolidated_files)
-    |> Enum.map(fn {_file, path} -> path |> to_charlist() end)
+    |> Enum.map(fn {_file, path} -> path end)
+    |> reject_excluded_files()
+    |> Enum.map(&to_charlist(&1))
+  end
+
+  defp reject_excluded_files(files) do
+    file_exclusions = dialyzer_config()[:excluded_file_suffixes] || []
+
+    Enum.reject(files, fn file ->
+      Enum.any?(file_exclusions, fn exclusion ->
+        String.match?(file, ~r/#{exclusion}$/)
+      end)
+    end)
   end
 
   defp dialyzer_paths do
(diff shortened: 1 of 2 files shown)


lib/dialyxir/project.ex
```diff
@@ -67,7 +67,19 @@ defmodule Dialyxir.Project do
...
+    |> reject_excluded_files()
+    |> Enum.map(&to_charlist(&1))
+  end
+
+  defp reject_excluded_files(files) do
+    file_exclusions = dialyzer_config()[:excluded_file_suffixes] || []
+
+    Enum.reject(files, fn file ->
+      Enum.any?(file_exclusions, fn exclusion ->
+        String.match?(file, ~r/#{exclusion}$/)
```
maltoe: https://hexdocs.pm/elixir/String.html#ends_with?/2 ?

juffel: Much easier on the eyes 👍

asummers:
I wonder if it would be better to have BEAM file match support for the ignore file instead. We support regexes there for the warnings already.

jeremyjh:
> I wonder if it would be better to have BEAM file match support for the ignore file instead. We support regexes there for the warnings already.

Its a good point - and actually you can already do that. 

 @juffel what is the motivation or use-case for this?

jeremyjh:
Also see https://github.com/jeremyjh/dialyxir/issues/264

juffel:
Thanks for the quick responses 💚

Might be that my use case for this is a little bit special. The project I want to have this feature for has some compile-optimizations, where certain files (e.g. files in `test/support`) are only compiled in certain environments (e.g. `test` environment). In consequence we got some errors on CI (which we configured dialyzer to run in `test` env) that we did not get locally in `dev` env and therefore we wanted to explicitly exclude these files in `test` env.

Now that I think about it again, that use case could also be solved by adding the respective warnings to the ignore list (since most of them had proper path references) or by simply running dialyzer in `test` env locally 🙈

However I imagine that it is a bit more handy to simply exclude certain files for which you don't care about warnings, instead of adding every single warning from them to the ignores file.
I can very well understand if you don't find this a valid use case, though @jeremyjh

jeremyjh:
@juffel You don't have to ignore every warning, you can ignore every warning from a specific path with a regex like in the issue I referenced above.  Can you try that first? The only downside I see is that you waste some time doing analysis on those files, but I'm guessing we're talking about milliseconds here for most people's test/support.

juffel:
True that should also work, closing this.

vans163:
@jeremyjh This is still relevent, the usecase is when you have very large modules (such as protobuf output) that dialyzer takes 30+ secs on per module, so you cached to the PLT and want to ignore it in the starting dialyzer files.

Issue #398, opened 2020-07-08
Exception on non-existent directory for :plt_core_path


jrmnl:
### Environment

elixir: 1.10.3
otp_release: 23.0

### Current behavior
When you set `:plt_core_path` to non existent folder you will get following exceiption on `mix dialyzer`:
```
...
Creating dialyxir_erlang-23.0.plt
:dialyzer.run error:     The PLT file ./my_non_existent_folder/dialyxir_erlang-23.0.plt is not writable
Looking up modules in dialyxir_erlang-23.0.plt
** (Mix) Could not open ./my_non_existent_folder/dialyxir_erlang-23.0.plt: no such file or directory
```

### Expected behavior
You can successfully create plt files in non existent folder.

jeremyjh:
> You can successfully create plt files in non existent folder.

Is this what people would really expect? Is there another mix task that handles non-existent paths that way for example? I'm just not really sure that the expectation should be to silently create the full path.

asummers:
I would support a `--create-if-not-exists` type flag but creating by default is surprising behavior to me as a user.

jrmnl:
> > You can successfully create plt files in non existent folder.
> 
> Is this what people would really expect? Is there another mix task that handles non-existent paths that way for example? I'm just not really sure that the expectation should be to silently create the full path.

If I specify `:plt_file` in non-existent path it will create it, for example. And here `:plt_core_path` sounds like output folder if it's empty or not exists.

`--create-if-not-exists` makes usage less obvious.

When this error was occured my first thoughts were:
1. Is provided path correct?
2. Do I have rights to create folders here?

jeremyjh:
@asummers after looking at this more closely I've kind of come around to @jrmnl point of view here - since this is a file we create, it may not be surprising to create its path and [we already do this](https://github.com/jeremyjh/dialyxir/blob/master/lib/dialyxir/plt.ex#L167) when we copy the base PLT's to the target directory.

PR #399, opened 2020-07-08
Fix non existent directory for :plt_core_path

jrmnl:
https://github.com/jeremyjh/dialyxir/issues/398


jeremyjh:
I don't remember why this was never merged. It would need a test for one thing. But I'm skeptical we need it.

PR #402, opened 2020-07-15
Remove nesting from warning and do not return self() on start

josevalim:
Erlang/OTP requires a separate supervisor process to
be returned from the start/2 callback. Otherwise this
may cause issues in tools like Observer, LiveDashboard,
etc.


jeremyjh:
Thanks!

PR #404, opened 2020-07-31
Test against more modern elixir and OTP versions

PragTob:

PR #406, opened 2020-08-04
add recent Elixir versions to CI

adrianomitre:


jeremyjh:
I think this is a dupe of #404 and it has the same problems in CI. At least the CI results are fixed though!

jeremyjh:
closed in #415

PR #421, opened 2020-12-23
Added a configuration option to specify the path of the project plt file

baseballlover723:
This is helpful if you want to move the dialyxir plt files into their own directory (in my case, this is helpful for running dialyxir inside of docker with caching), but don't want to mess with the naming of the plt files.


jeremyjh:
This is a good change but can you please add it to the README and Module Docs in the appropriate places?

baseballlover723:
@jeremyjh updated

jeremyjh (approved):

PR #427, opened 2021-03-21
Use Github Actions instead of TravisCI

dolfinus:
Hello.

I've realized that there are no tests running at commit/PR into your repo.
It seems that you've previously used TravisCI for that. I'm not sure why is it disabled now, but I personally faced the issue with their new billing plans and too strict limitations for running builds for open source projects. Maybe this is your case too.

Here I've replaced TravisCI integration with Github Actions. All steps remain the same. You can check results here:
https://github.com/dolfinus/dialyxir/runs/2159686660

What do you think about this PR?


jeremyjh (approved):

jeremyjh:
Hi, thank you! Yes I'd gotten an email that Travis was going away and have been meaning to do this, so I really appreciate you jumping in and taking care of it!

PR #436, opened 2021-08-03
update CI to the latest Elixir 1.11 and Erlang 23. Add Elixir 1.12 and Erlang 24

baseballlover723:
Also upgrades the local Elixir version to `1.12.2` and the Erlang version to `24.0.5`. This also includes a `mix format` since Elixir 1.12 made a backwards incompatible formatting change as well as only running the format check in CI on Elixir 1.12+. 

This also changes CI to not fail as soon a single job fails, this is needed for now, since CI fails on Erlang 24. This is probably for the better since you can see if CI fails for just a single Elixir / Erlang combination or for all of them.


jeremyjh:
Hi, thanks for working on this! I think to resolve the errors on OTP 24 we could just remove this entire check from the tests: https://github.com/jeremyjh/dialyxir/blob/242c0e89fd08535eedfdb971993bb550b32bdcce/.github/workflows/test.yml#L97; it would be hard to keep broad compatibility and still check this.

Also, I would be fine with dropping the 1.9.4 targets entirely since that is 4 releases ago, my original thought was to support the 3 most recent releases anyway.

baseballlover723:
This has been updated

baseballlover723:
I have no idea why the tests fail now, but most of them did work when I originally pushed it up, https://github.com/jeremyjh/dialyxir/runs/3248958978

jeremyjh:
The problem is we're no longer running any dialyzer check before the output_test, so output_test has to build the PLTs etc. Re-ordering the output tests [like this](https://github.com/jeremyjh/dialyxir/pull/440/commits/9acf8808e276220e4fbe10896f7391c164e02b7e) will fix it.

jeremyjh (approved):

jeremyjh:
Thanks!

PR #441, opened 2021-09-10
only run CI on master and on PRs against master

baseballlover723:
This is slightly more constrained and will also still work if you decide to change the default branch

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
+    branches: [ master ]
   pull_request:
-    branches:
-      - master
+    branches: [ master ]
 
 jobs:
   check_duplicate_runs:


jeremyjh (approved):

jeremyjh:
It looks like it didn't actually run any checks though?

baseballlover723:
It seems like this `$default-branch` is only for workflow templates, and not workflows (it hardcodes it once you initialize a workflow from a template), it seems like we could use something like https://stackoverflow.com/questions/64781462/github-actions-default-branch-variable though, let me give it a try

baseballlover723:
seems like that didn't work either, i'm just going to revert it back to using the hardcoded master

baseballlover723:
as a side note, what are your thoughts on removing the duplicate runs check? my main concern is for instance the PR that was merged earlier, the commit looks like https://github.com/jeremyjh/dialyxir/commit/e8c46d61589d5b7b2878856b930da191c767b64e, and if you hover over the CI check, it basically says that it was skipped. Now you could go and open up the PR and see the checks there, but I think it would be clearer if we just ran CI always, even if its duplicated. Though this would imply running more github action jobs, which I don't really know if that actually makes a difference.

jeremyjh:
Thanks! I think it is ok if CI on the commit on master is skipped as there is a branch protection rule so that PRs are not merged unless the checks pass.

PR #463, opened 2022-07-12
feat: Adds a github formatter

isaacsanders:
I would am working to create a github actions formatter for this library, and the first step I wanted to take is to extract the other formats into modules.

### DIFF
--- a/lib/dialyxir/formatter/github.ex
+++ b/lib/dialyxir/formatter/github.ex
@@ -4,12 +4,22 @@ defmodule Dialyxir.Formatter.Github do
   @behaviour Dialyxir.Formatter
 
   @impl Dialyxir.Formatter
-  def format({_tag, {file, line}, {warning_name, _arguments}}) do
+  def format({_tag, {file, line}, {warning_name, arguments}}) do
     base_name = Path.relative_to_cwd(file)
 
     warning = warning(warning_name)
     string = warning.format_short(arguments)
 
     "::warning file=#{base_name},line=#{line},title=#{warning_name}::#{string}"
+  end
+
+  defp warning(warning_name) do
+    warnings = Dialyxir.Warnings.warnings()
+
+    if Map.has_key?(warnings, warning_name) do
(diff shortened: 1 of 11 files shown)


jeremyjh:
Hi, I'm not opposed to the refactoring but can you explain a bit more about what you are trying to accomplish? Is it returning the results in a structure understood by github actions so they appear better in PRs? And then you'll do another PR to implement an action formatter?

isaacsanders:
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message

I have the action formatter written already, but I figured doing changes in smaller chunks was a better approach.

The link above outlines the format. It allows for inline annotations of the files from the github check run, i believe.

isaacsanders:
Added in. Thanks for the review and promptness!

lib/dialyxir/formatter/github.ex
```diff
@@ -0,0 +1,15 @@
...
+  @behaviour Dialyxir.Formatter
+
+  @impl Dialyxir.Formatter
+  def format({_tag, {file, line}, {warning_name, _arguments}}) do
+    base_name = Path.relative_to_cwd(file)
+
+    warning = warning(warning_name)
+    string = warning.format_short(arguments)
+
+    "::error file=#{base_name},line=#{line},title=#{warning_name}::#{string}"
```
isaacsanders, line 13: @jeremyjh Would you say that everything raised is an error?

isaacsanders: I've moved it down to "warning"

isaacsanders:
I've added the docs

jeremyjh (approved):
Looks great, thank you!

jeremyjh:
@isaacsanders I just realized I had to release it to run CI, there are a couple of compile warnings can you please check?

jeremyjh:
I've published to hex as 1.2.

isaacsanders:
Thanks Jeremy! For this and for being a maintainer!

PR #473, opened 2022-11-06
Fix crash when mix.lock was missing

### LINKED ISSUE #472: Crashes if `mix.lock` does not exist
### Precheck

- [x] Take a look at the open issues and be sure that your issue is not already covered.
- [x] Be sure your versions of Dialyxir and Erlex are up to date.

### Environment

* Elixir & Erlang/OTP versions (elixir --version):

```
Erlang/OTP 24 [erts-12.3.2.6] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [jit]

Elixir 1.14.1 (compiled with Erlang/OTP 24)
```

* Which version of Dialyxir are you using? (cat mix.lock | grep dialyxir): 1.2.0

### Current behavior
I have `mix dialyzer` installed globally as an archive task. If I make a new project with `mix new`, and then run `mix dialyzer`, `mix crashes with 

```
$ mix dialyzer
Finding suitable PLTs
Checking PLT...
[:compiler, :elixir, :kernel, :logger, :stdlib]
** (File.Error) could not read file "mix.lock": no such file or directory
    (elixir 1.14.1) lib/file.ex:358: File.read!/1
    lib/mix/tasks/dialyzer.ex:385: Mix.Tasks.Dialyzer.dependency_hash/0
    lib/mix/tasks/dialyzer.ex:249: Mix.Tasks.Dialyzer.check_plt/1
    lib/mix/tasks/dialyzer.ex:174: Mix.Tasks.Dialyzer.run/1
    (mix 1.14.1) lib/mix/task.ex:421: anonymous fn/3 in Mix.Task.run_task/4
    (mix 1.14.1) lib/mix/cli.ex:84: Mix.CLI.run_task/2
```

It looks like if I create a `mix.lock` file containing `%{}`, then it runs as expected.

### Expected behavior
I'd expect it not to crash with an empty `mix.lock`.

ollien:
Closes #472

See that issue for more details on the issue. A non-existent lockfile is probably ok to be non-fatal (especially considering by default `mix` projects do not include one), so I just chose to treat it the same as an empty file.

### DIFF
--- a/lib/mix/tasks/dialyzer.ex
+++ b/lib/mix/tasks/dialyzer.ex
@@ -386,7 +386,23 @@ defmodule Mix.Tasks.Dialyzer do
     {apps, hash}
   end
 
-  def lock_file() do
-    Mix.Project.config()[:lockfile] |> File.read!()
+  defp lock_file() do
+    lockfile = Mix.Project.config()[:lockfile]
+    read_res = File.read(lockfile)
+
+    case read_res do
+      {:ok, data} ->
+        data
+
+      {:error, :enoent} ->
+        # If there is no lock file, an empty bitstring will do to indicate there is none there
+        <<>>
+
+      {:error, reason} ->
+        raise File.Error,
+          reason: reason,
+          action: "read file",
+          path: lockfile
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


jeremyjh (approved):

jeremyjh:
Thanks!

PR #481, opened 2023-03-10
Add `:no-umbrella` config option

pnezis:
In some cases you may want to place the `lockfile` in a parent folder without having an umbrella. With this extra option you avoid compilation at the parent folder which will fail since in such cases there is no `mix.exs`. If this flag is not set then the behaviour remains unchanged.


pnezis:
@jeremyjh you are right, updated the PR accordingly

jeremyjh (approved):

PR #491, opened 2023-04-14
Add CircleCI example configuration

merwan:
Creating the cache key from the Elixir and Erlang version was non-trivial, this will probably be useful for other people too.

### DIFF
--- a/docs/circleci.md
+++ b/docs/circleci.md
@@ -1,0 +1,40 @@
+# CircleCI
+
+```yaml
+---
+version: 2
+
+jobs:
+  build:
+    docker:
+      - image: cimg/elixir:1.14
+
+    steps:
+      - checkout
+
+      # Compile steps omitted for simplicity
+
+      # Don't cache PLTs based on mix.lock hash, as Dialyzer can incrementally update even old ones
+      # Cache key based on Elixir & Erlang version (also useful when running in matrix)
+      - run:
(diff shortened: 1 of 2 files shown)


docs/circleci.md
```diff
@@ -0,0 +1,40 @@
...
+      # Don't cache PLTs based on mix.lock hash, as Dialyzer can incrementally update even old ones
+      # Cache key based on Elixir & Erlang version (also useful when running in matrix)
+      - run:
+          name: "Save Elixir and Erlang version for PLT caching"
+          command: echo "$ELIXIR_VERSION $ERLANG_VERSION" > .elixir_otp_version
+
+      - restore_cache:
+          name: "Restore PLT cache"
+          keys:
+            - {{ arch }}-{{ checksum ".elixir_otp_version" }}-plt
```
jeremyjh, line 26: In my app I just do a `{{ checksum ".circleci/config.yml" }}`. This works in a file like this one because the image tag will change when the Elixir/Erlang version changes.

merwan: Our CircleCI configuration is much more complicated as we have many other steps and jobs in the workflow. We don't want to invalidate the cache when the file changes because there can be many other modifications unrelated to the Erlang or Elixir version.
I can provide a more complete CircleCI configuration example if needed.

jeremyjh (approved):

jeremyjh:
Thanks!

Issue #492, opened 2023-04-14
Format option for ignore file and output: {file, short_description}


Nezteb:
According to [the docs about Elixir term format](https://github.com/jeremyjh/dialyxir#elixir-term-format), the following options are supported for an ignore file:
```elixir
# .dialyzer_ignore.exs
[
  # {short_description}
  {":0:unknown_function Function :erl_types.t_is_opaque/1/1 does not exist."},
  # {short_description, warning_type}
  {":0:unknown_function Function :erl_types.t_to_string/1 does not exist.", :unknown_function},
  # {short_description, warning_type, line}
  {":0:unknown_function Function :erl_types.t_to_string/1 does not exist.", :unknown_function, 0},
  # {file, warning_type, line}
  {"lib/dialyxir/pretty_print.ex", :no_return, 100},
  # {file, warning_type}
  {"lib/dialyxir/warning_helpers.ex", :no_return},
  # {file}
  {"lib/dialyxir/warnings/app_call.ex"},
  # regex
  ~r/my_file\.ex.*my_function.*no local return/
]
```

`mix dialyzer --format ignore_file` returns a list of warnings using the `{file, warning_type}`, which is very useful.

I'd love to go one step further and generate a list of warnings with the format `{file, short_description}`. If a single file has more than one instance of a warning, a single line in the ignore file will end up ignoring all of the warnings. Sometimes I'd prefer to ignore each individual warning specifically.

For example, if `mix dialyzer --format short` gives me:
```
lib/something.ex:15:no_return Function init/1 has no local return.
lib/something.ex:26:no_return Function update/1 has no local return.
lib/something.ex:36:no_return Function refresh/1 has no local return.
lib/something.ex:45:no_return Function update_state_from_cache/2 has no local return.
lib/something.ex:49:no_return Function update_state_from_backend/2 has no local return.
```
Then this single ignore line will ignore all five of those warnings:
```elixir
# .dialyzer_ignore.exs
[
  # {file, warning_type}
  {"lib/something.ex", :no_return},
]
```

I'd like to be able to do:
```elixir
# .dialyzer_ignore.exs
[
  # {file, short_description}
  {"lib/something.ex", "Function init/1 has no local return."},
  {"lib/something.ex", "Function update/1 has no local return."},
  {"lib/something.ex", "Function refresh/1 has no local return."},
  {"lib/something.ex", "Function update_state_from_cache/2 has no local return."},
[...]

PR #493, opened 2023-04-14
Add new format for ignore_file_strict: {file, warning_description}

Nezteb:
Resolves https://github.com/jeremyjh/dialyxir/issues/492

### DIFF
--- a/lib/dialyxir/project.ex
+++ b/lib/dialyxir/project.ex
@@ -117,7 +117,22 @@ defmodule Dialyxir.Project do
   end
 
   defp skip?({file, warning, line}, {file, warning, line, _}), do: true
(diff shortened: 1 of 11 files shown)


lib/dialyxir/project.ex
```diff
@@ -117,7 +117,22 @@ defmodule Dialyxir.Project do
...
+    description =
+      description
+      |> String.split(":")
+      |> Enum.at(2)
+      |> String.split(" ")
+      |> Enum.slice(1..-1)
+      |> Enum.join(" ")
+
+    short_description == description
+  end
```
Nezteb, line 133: @jeremyjh Thoughts on a better way to do this? It works but is ugly. 😅

jeremyjh:
Thanks!

PR #505, opened 2023-06-13
Fix gitlab_ci cache files key example

wkirschbaum:
The current example is invalid:
jobs:build-dev:cache config contains unknown keys: files

### DIFF
--- a/docs/gitlab_ci.md
+++ b/docs/gitlab_ci.md
@@ -20,9 +20,9 @@ build-dev:
   stage: compile
   cache:
     - key:
-      files:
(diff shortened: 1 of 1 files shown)


jeremyjh:
> I also just want to note that by default plt's will sit in the _build/dev folder, not the priv folder. Does it make sense to perhaps change this?

Well these are essentially build artifacts, and we don't want them checked into git so it makes sense to me to put them in the `_build` folder which is .gitignored by default, in `.priv` you'd have to `.gitignore` them particularly.

PR #512, opened 2023-07-19
Elixir 1.15

### LINKED ISSUE #511: `mix dialyzer` reports dependency missing
### Environment

* Elixir & Erlang/OTP versions (elixir --version):
```
Erlang/OTP 26 [erts-14.0.2] [source] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1] [jit]
Elixir 1.15.1 (compiled with Erlang/OTP 24)
```

* Which version of Dialyxir are you using? (cat mix.lock | grep dialyxir): `"dialyxir": {:hex, :dialyxir, "1.3.0", "fd1672f0922b7648ff9ce7b1b26fcf0ef56dda964a459892ad15f6b4410b5284", [:mix], [{:erlex, ">= 0.2.6", [hex: :erlex, repo: "hexpm", optional: false]}], "hexpm", "00b2a4bcd6aa8db9dcb0b38c1225b7277dca9bc370b6438715667071a304696f"},`

### Current behavior

`mix dialyzer` in this project prints

```
DEPENDENCY MISSING
------------------------
If you are reading this message, then Elixir and Erlang are installed but the
Erlang Dialyzer is not available. Probably this is because you installed Erlang
with your OS package manager and the Dialyzer package is separate.

On Debian/Ubuntu:

  `apt-get install erlang-dialyzer`

Fedora:

   `yum install erlang-dialyzer`

Arch and Homebrew include Dialyzer in their base erlang packages. Please report a Github
issue to add or correct distribution-specific information.
```

but using e.g. 

```
$ iex -S mix
Erlang/OTP 26 [erts-14.0.2] [source] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1] [jit]

Interactive Elixir (1.15.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> Code.ensure_loaded?(:dialyzer)
true
iex(2)>
```

The elixir/erlang is installed on macOS via asdf, `.tool-versions`:

```
elixir 1.15.1
erlang 26.0.2
```

It's an umbrella project and I run it in the umbrella itself.

### Expected behavior

After I upgraded dialyxir, elixir and erlang I expected it to work the same as before.

danschultzer:
Resolves #511 

This uses `Mix.ensure_application!/1` instead of `Code.ensure_loaded?/1`. This is similar to how missing erlang libraries are dealt with in other repos:

https://github.com/nccgroup/sobelow/pull/143
https://github.com/phoenixframework/tailwind/commit/68185c77c7f83c6b634cc7624b19633cb61b83d2

This PR also tests on Elixir 1.15 and OTP 26, but there are issues getting the tests to succeed. I'll see if I can get the tests to succeed.

### DIFF
--- a/lib/mix/tasks/dialyzer.ex
+++ b/lib/mix/tasks/dialyzer.ex
@@ -339,28 +339,34 @@ defmodule Mix.Tasks.Dialyzer do
     end
   end
 
-  defp check_dialyzer do
-    if not Code.ensure_loaded?(:dialyzer) do
-      error("""
-      DEPENDENCY MISSING
-      ------------------------
-      If you are reading this message, then Elixir and Erlang are installed but the
-      Erlang Dialyzer is not available. Probably this is because you installed Erlang
-      with your OS package manager and the Dialyzer package is separate.
+  if Version.match?(System.version(), ">= 1.15.0") do
+    defp check_dialyzer do
+      Mix.ensure_application!(:dialyzer)
+    end
+  else
+    defp check_dialyzer do
+      if not Code.ensure_loaded?(:dialyzer) do
+        error("""
+        DEPENDENCY MISSING
+        ------------------------
+        If you are reading this message, then Elixir and Erlang are installed but the
+        Erlang Dialyzer is not available. Probably this is because you installed Erlang
+        with your OS package manager and the Dialyzer package is separate.
 
-      On Debian/Ubuntu:
+        On Debian/Ubuntu:
 
-        `apt-get install erlang-dialyzer`
+          `apt-get install erlang-dialyzer`
 
-      Fedora:
+        Fedora:
 
-         `yum install erlang-dialyzer`
+          `yum install erlang-dialyzer`
 
-      Arch and Homebrew include Dialyzer in their base erlang packages. Please report a Github
-      issue to add or correct distribution-specific information.
-      """)
+        Arch and Homebrew include Dialyzer in their base erlang packages. Please report a Github
+        issue to add or correct distribution-specific information.
+        """)
 
-      :erlang.halt(3)
+        :erlang.halt(3)
+      end
     end
   end
 
--- a/test/fixtures/local_plt_no_warn/mix.exs
+++ b/test/fixtures/local_plt_no_warn/mix.exs
@@ -2,6 +2,11 @@ defmodule LocalPltNoWarn.Mixfile do
   use Mix.Project
 
   def project do
-    [app: :local_plt_no_warn, version: "1.0.0", dialyzer: [plt_file: {:no_warn, "local.plt"}]]
+    [
+      app: :local_plt_no_warn,
+      prune_code_paths: false,
+      version: "1.0.0",
+      dialyzer: [plt_file: {:no_warn, "local.plt"}]
+    ]
   end
 end
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,14 +37,28 @@ jobs:
         - '1.12.3'
         - '1.13.4'
         - '1.14.3'
+        - '1.15.4'
         otp:
         - '23.3'
         - '24.3'
         - '25.3'
+        - '26.0'
 
         exclude:
           - elixir: '1.12.3'
             otp: '25.3'
+
+          - elixir: '1.12.3'
+            otp: '26.0'
+
+          - elixir: '1.13.4'
+            otp: '26.0'
+
+          - elixir: '1.14.3'
+            otp: '26.0'
+
+          - elixir: '1.15.4'
+            otp: '23.3'
 
     steps:
     - name: Checkout
@@ -87,7 +101,7 @@ jobs:
         MIX_ENV: prod
 
     - name: Check source code formatting
-      if: ${{ matrix.elixir > '1.12.0' }}
+      if: ${{ matrix.elixir > '1.15.0' }}
       run: mix format --check-formatted
 
     - name: Get results in short format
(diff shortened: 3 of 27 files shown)


danschultzer:
@jeremyjh it would be helpful to me if you can approve the workflow so the CI can run.  I have to switch in local right now to test between them.

jeremyjh:
I changed the workflow settings so it shouldn't prompt again.

jonnystorm:
@danschultzer FYI: I just got around to engaging those errors for my own [project](https://gitlab.com/jonnystorm/gitlab-elixir-plts). For what it's worth, I cleaned up tests with [this commit](https://github.com/jonnystorm/dialyxir/commit/92600f57c48572f5ca34101cf7d3ec955cedf7de). Some additions are reasonable while others are clearly kludges, but none of the added `require`s, etc. are superfluous (that I can see), and none can be pushed down deeper into the dependency tree.

I plan to take a similar inventory of the various errors observed while running as an archive with `mix`, but feel free to take whatever is useful from the above and throw away the rest.

danschultzer:
Thanks @jonnystorm! I noticed that the dialyxir load paths disappears in certain cases using `:code.get_path`, and this only happens after the `Mix.Project.in_project` calls. So I've set `prune_code_paths: false` on all fixtures.

This may break in the future and I would rather ensure that the `Mix.Project.in_project` callback doesn't cause the dialyxir path to be unloaded. I couldn't grok how to fix that. I guess all fixtures has to be updated to load the dialyxir dependency, but I couldn't get it to work properly.

@jeremyjh this should be ready for review!

elishagreenwald:
Hi is there a plan to merge this soon? Wondering if I should explore a workaround / hold off on upgrading etc... any timelines would be super helpful. Thanks for all the great work folks!

lib/mix/tasks/dialyzer.ex
```diff
@@ -339,28 +339,36 @@ defmodule Mix.Tasks.Dialyzer do
...
-      error("""
-      DEPENDENCY MISSING
-      ------------------------
-      If you are reading this message, then Elixir and Erlang are installed but the
-      Erlang Dialyzer is not available. Probably this is because you installed Erlang
-      with your OS package manager and the Dialyzer package is separate.
+  if Version.match?(System.version(), ">= 1.15.0") do
+    defp check_dialyzer do
+      if function_exported?(Mix, :ensure_application!, 1) do
+        Mix.ensure_application!(:dialyzer)
```
jeremyjh, line 345: If this raises - for example because they installed erlang on debian without dialyzer - is it obvious what the user needs to do next? That is why we are printing the extended error message in the other branch of this top-level if. I think we should trap this and print the message in this case as well, maybe just move the message itself to separate function.

jeremyjh:
@danschultzer @elishagreenwald Hi guys, sorry to leave this languishing, I'd somehow missed it was complete. Been super busy lately. I have just a couple of small asks to change before we merge.

test/fixtures/local_plt_no_warn/mix.exs
```diff
@@ -2,6 +2,11 @@ defmodule LocalPltNoWarn.Mixfile do
   use Mix.Project
 
   def project do
-    [app: :local_plt_no_warn, version: "1.0.0", dialyzer: [plt_file: {:no_warn, "local.plt"}]]
+    [
+      app: :local_plt_no_warn,
+      prune_code_paths: false,
```
jeremyjh, line 7: prune_cod_paths is a workaround for apps that aren't specifying all their runtime dependencies. Could add dialyzer to extra_applications instead? I don't think `prune_code_paths` is what we want to recommend people do.

lib/mix/tasks/dialyzer.ex
danschultzer: This is the message printed: https://github.com/elixir-lang/elixir/blob/a94700f861706482f6fe7f5abb884f235763db6f/lib/mix/lib/mix.ex#L622-L627

Which would read as:

> The application "dialyzer" could not be found. This may happen if your
> Operating System broke Erlang into multiple packages and may be fixed 
> by installing the missing "erlang-dev" and "erlang-dialyzer" packages

It should be clear enough right?

lib/mix/tasks/dialyzer.ex
```diff
@@ -339,28 +339,36 @@ defmodule Mix.Tasks.Dialyzer do
...
-    if not Code.ensure_loaded?(:dialyzer) do
-      error("""
-      DEPENDENCY MISSING
-      ------------------------
-      If you are reading this message, then Elixir and Erlang are installed but the
-      Erlang Dialyzer is not available. Probably this is because you installed Erlang
-      with your OS package manager and the Dialyzer package is separate.
+  if Version.match?(System.version(), ">= 1.15.0") do
+    defp check_dialyzer do
+      if function_exported?(Mix, :ensure_application!, 1) do
```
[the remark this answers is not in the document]
danschultzer: Good catch, fixed.

test/fixtures/local_plt_no_warn/mix.exs
danschultzer: It's Dialyxir that [gets pruned](https://github.com/jeremyjh/dialyxir/pull/512#issuecomment-1646697518), and I haven't been able to find a way to prevent Elixir from pruning it. Halfway through the tests all the Dialixir modules are gone. I've tried `extra_applications`, and setting it as a dep. Only thing I've found to work is to setting that flag.

This could break in the future, so it would be nice to find a better solution, but I haven't figured out why `Mix.Project.in_project` would cause the dialixir code paths to be pruned in the first place.

jeremyjh (approved):

PR #509, opened 2023-07-06
Do not load optional applications

aptinio:
Thanks to @josevalim's explanation:
https://elixirforum.com/t/elixir-v1-15-0-rc-0-released/56019/27

Fixes #502.

Issue #502, opened 2023-05-26
Dialyxir attempts to load transitive optional dependencies


kipcole9:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.15.0-rc.0, OTP 26.0

* Which version of Dialyxir are you using? (cat mix.lock | grep dialyxir): 1.3.0

### Current behavior

Dialyxir is issuing warnings that it can't load applications that are transitive optional dependencies. An example using the [ex_cldr](https://github.com/elixir-cldr/cldr) lib:
```elixir
kip@Kips-MacBook-Pro cldr % asdf current
elixir          1.15.0-rc.0-otp-26 /Users/kip/Development/cldr/.tool-versions
erlang          26.0            /Users/kip/Development/cldr/.tool-versions
rust            stable          /Users/kip/.tool-versions
kip@Kips-MacBook-Pro cldr % mix dialyzer
Finding suitable PLTs
Checking PLT...
Error loading castore, dependency list may be incomplete.
 {~c"no such file or directory", ~c"castore.app"}
Error loading certifi, dependency list may be incomplete.
 {~c"no such file or directory", ~c"certifi.app"}
[:asn1, :castore, :certifi, :cldr_utils, :compiler, :crypto, :decimal, :earmark_parser, :eex, :elixir, :ex_doc, :ex_unit, :expo, :gettext, :inets, :jason, :kernel, :logger, :makeup, :makeup_elixir, :makeup_erlang, :mix, :nimble_parsec, :public_key, :ssl, :stdlib, :sweet_xml, :xmerl]
PLT is up to date!
ignore_warnings: .dialyzer_ignore_warnings

Starting Dialyzer
[
  check_plt: false,
  init_plt: ~c"/Users/kip/Development/cldr/_build/dev/dialyxir_erlang-26.0_elixir-1.15.0-rc.0_deps-dev.plt",
  files: [~c"/Users/kip/Development/cldr/_build/dev/lib/ex_cldr/ebin/Elixir.Cldr.LanguageTag.Chars.Cldr.LanguageTag.U.beam",
   ~c"/Users/kip/Development/cldr/_build/dev/lib/ex_cldr/ebin/Elixir.Cldr.Normalize.CalendarEra.beam",
   ~c"/Users/kip/Development/cldr/_build/dev/lib/ex_cldr/ebin/Elixir.Cldr.NoParentError.beam",
[...]

PR #514, opened 2023-08-16
Fixed an issue related to missing optional applications

GRoguelon:
The PR updates the deps_app/2 function in the Dialyxir.Project module in `lib/dialyxir/project.ex` to handle different versions of Elixir. Added version checking for Elixir 1.15 or later, and adjusted the implementation accordingly. The changes ensure that the `deps_app/2` function works correctly with the appropriate version of Elixir.

I'm not totally sure if my change is backward compatible with previous versions of Elixir. Also not sure if there is a better option to avoid the version check.

I used this patch for an Elixir/Phoenix application on 1.15 for several weeks without any issues and no warnings about optional applications.

Closes #509

### DIFF
--- a/lib/dialyxir/project.ex
+++ b/lib/dialyxir/project.ex
@@ -362,38 +362,84 @@ defmodule Dialyxir.Project do
     deps_app(app, recursive)
   end
 
-  @spec deps_app(atom(), boolean()) :: [atom]
-  defp deps_app(app, recursive) do
-    with_each =
-      if recursive do
-        &deps_app(&1, true)
-      else
-        fn _ -> [] end
+  if System.version() |> Version.parse!() |> then(&(&1.major >= 1 and &1.minor >= 15)) do
+    @spec deps_app(atom(), boolean()) :: [atom()]
+    defp deps_app(app, recursive) do
+      case do_load_app(app) do
+        :ok ->
+          with_each =
+            if recursive do
+              &deps_app(&1, true)
+            else
+              fn _ -> [] end
+            end
+
+          # Identify the optional applications which can't be loaded and thus not available
+          missing_apps =
+            Application.spec(app, :optional_applications)
+            |> List.wrap()
+            |> Enum.reject(&(do_load_app(&1) == :ok))
+
+          # Remove the optional applications which are not available from all the applications
+          required_apps =
+            Application.spec(app, :applications)
+            |> List.wrap()
+            |> Enum.reject(&(&1 in missing_apps))
+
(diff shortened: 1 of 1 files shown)


axelson:
Thanks for this! I tested it on a work repository and it works as advertised :+1:

This should be marked as Fixes https://github.com/jeremyjh/dialyxir/issues/502 as well.

jeremyjh (approved):

jeremyjh:
Thanks!

whatyouhide:
@jeremyjh @GRoguelon this inadvertently bumps the Elixir requirement to 1.14+ because it uses `then/2`. @GRoguelon maybe you wanna send a PR that doesn't use `then/2`, so that this stays compatible with Elixir 1.6+, like `mix.exs` declares? 🙃

whatyouhide:
@jeremyjh what I’m saying is that if the library is **only** compatible with Elixir 1.15+, then we should change the requirement in `mix.exs` ([here](https://github.com/jeremyjh/dialyxir/blob/7285f152418def3dd28c318f49613054f3c7e788/mix.exs#L11)) to `~> 1.12`. I see that you did that on the main branch, but the problem is that the 1.4.0 release [declares it supports `1.6+`](https://preview.hex.pm/preview/dialyxir/1.4.0/show/mix.exs#L11), but it uses [`then/2`](https://hexdocs.pm/elixir/Kernel.html#then/2), which was released in Elixir 1.12.

So right now it might be just a matter of releasing 1.4.1 as a bug fix 😉

jeremyjh:
I released 1.4.1 last night.

bryannaegele:
It would be nice if the change could be updated to not use `then` as while this project may only be testing 3 versions back, we have to support back further for some major libraries as we base off of supported OTP version rather than Elixir. We still have to support back to 1.11 for another year (OTP 27).

David-Klemenc:
Something like this:

```
== Compilation error in file lib/dialyxir/project.ex ==
Error: ** (CompileError) lib/dialyxir/project.ex:365: undefined function then/2
```

```elixir
  @spec check_version(Version.t()) :: boolean()
  defp check_version(%Version{major: major, minor: minor}) do
    major >= 1 and minor >= 15
  end

  if System.version() |> Version.parse!() |> check_version() do
```

David-Klemenc:
Something like this:

```
== Compilation error in file lib/dialyxir/project.ex ==
Error: ** (CompileError) lib/dialyxir/project.ex:365: undefined function then/2
```

```elixir
if System.version() |> Version.parse!() |> (&(&1.major >= 1 and &1.minor >= 15)).() do
```

Issue #508, opened 2023-06-28
Elixir 1.15 + Umbrella Project + Dialyzer issues


elliottneilclark:
### Environment

* Elixir & Erlang/OTP versions (elixir --version):

```
Elixir 1.15.0 (compiled with Erlang/OTP 25)
```

* Which version of Dialyxir are you using? (cat mix.lock | grep dialyxir):
```
  "dialyxir": {:hex, :dialyxir, "1.3.0", "fd1672f0922b7648ff9ce7b1b26fcf0ef56dda964a459892ad15f6b4410b5284", [:mix], [{:erlex, ">= 0.2.6", [hex: :erlex, repo: "hexpm", optional: false]}], "hexpm", "00b2a4bcd6aa8db9dcb0b38c1225b7277dca9bc370b6438715667071a304696f"},
```

### Current behavior

I have an umbrella project with Phoenix + LiveView that worked well on 1.14 and is broken in 1.15. To reproduce it's pretty easy:

- mix phx.new --umbrella example
- add ` {:dialyxir, "~> 1.0", only: [:dev, :test], runtime: false}` to the deps
- `mix do deps.get, compile`
- `mix dialyzer`

```
Looking up modules in dialyxir_erlang-25.3.2.2_elixir-1.15.0_deps-dev.plt
** (UndefinedFunctionError) function :dialyzer_plt.included_files/1 is undefined (module :dialyzer_plt is not available)
    (dialyzer 5.0.5) :dialyzer_plt.included_files(~c"/home/elliott/Code/dialyzer_new_umbrella/_build/dev/dialyxir_erlang-25.3.2.2_elixir-1.15.0_deps-dev.plt")
    (dialyzer 5.0.5) dialyzer.erl:228: :dialyzer.plt_info/1
    (dialyxir 1.3.0) lib/dialyxir/plt.ex:243: Dialyxir.Plt.plt_files/1
    (dialyxir 1.3.0) lib/dialyxir/plt.ex:16: Dialyxir.Plt.find_plts/3
    (dialyxir 1.3.0) lib/mix/tasks/dialyzer.ex:254: Mix.Tasks.Dialyzer.check_plt/1
    (dialyxir 1.3.0) lib/mix/tasks/dialyzer.ex:174: Mix.Tasks.Dialyzer.run/1
    (mix 1.15.0) lib/mix/task.ex:447: anonymous fn/3 in Mix.Task.run_task/5
[...]

PR #518, opened 2023-09-16
Improve performance of algorithm to determine project deps

dhedlund:
Introduces smarter tracking of deps as they are loaded. These changes help to avoid traversing the same dep sub-trees multiple times. For larger projects, including umbrella projects, the same dep can be traversed millions of times.

One of our larger projects is an umbrella project with 20 umbrella apps and a little over 200 deps in our mix.lock file. With Elixir 1.15, the time it takes for dialyzer to complete increased 3x due to this issue, from 15 minutes in our CI to over 45 minutes. In order to get dialyxir working on Elixir 1.15, I also had to follow the advice in https://github.com/jeremyjh/dialyxir/issues/508 to add the dialyzer dep to each app in the umbrella. I'm not sure if that caused the increase in deps being returned, or if it's due to the new logic around optional apps.

The following table represents the time it takes to run the `Dialyxir.Project.include_deps/0` function against that project on my local (faster than CI machine), and how many apps the function returns:

| Elixir / OTP Version | Branch | Execution Time | Num Apps Returned |
|--------|--------|--------|--------|
| elixir1.14.4-otp-25 / erlang 25.3.1 | master | 33.6s | 16,845,380 |
| elixir1.14.4-otp-25 / erlang 25.3.1 | deps-tree-perf-improvements | 0.02s | 224 |
| elixir 1.15.5-otp-26 / erlang 26.0.2 | master | 15m 51.8s | 44,159,867 |
| elixir 1.15.5-otp-26 / erlang 26.0.2 | deps-tree-perf-improvements | 0.19s | 224 |

### DIFF
--- a/lib/dialyxir/project.ex
+++ b/lib/dialyxir/project.ex
@@ -6,6 +6,9 @@ defmodule Dialyxir.Project do
   alias Dialyxir.Formatter.Short
   alias Dialyxir.Formatter.Utils
 
+  # Maximum depth in the dependency tree to traverse before giving up.
+  @max_dep_traversal_depth 100
+
   def plts_list(deps, include_project \\ true, exclude_core \\ false) do
     elixir_apps = [:elixir]
     erlang_apps = [:erts, :kernel, :stdlib, :crypto]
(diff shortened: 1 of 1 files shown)


jeremyjh (approved):

jeremyjh:
Thanks!

PR #543, opened 2024-09-26
Add a check whether default_ignore file is empty during execution and ignores it if is

lnenad:
fixes #405 #542

Right now if the default ignore file `.dialyzer_ignore.exs` is empty the task will bug out as per issue https://github.com/jeremyjh/dialyxir/issues/542. This PR adds a check for file size, in case it is 0 it proceeds as if the file didn't exist and prints out an info message explaining that the file is empty.

`:ignore_warnings opt specified in mix.exs: .dialyzer_ignore.exs, but file is empty.`

Unfortunately, without my changes, the test suite is failing so I wasn't sure about adding a test for this, also my lack of in depth testing experience with Elixir.

PR #545, opened 2024-09-30
Fix issue with expecting default ignore file when checking file size 

lnenad:
This PR fixes issue with expecting default ignore file when checking file size and adds case to check custom ignore file size. The issue was made by my previous PR https://github.com/jeremyjh/dialyxir/pull/543 due to the fact it was always trying to `File.stat!` the default ignore file. 

I've manually tested the following 5 combinations: 

* no ignore file at all no config
* default ignore file no config not empty
* default ignore file no config empty
* non-default ignore file path in config not empty
* non-default ignore file path in config empty

They all work but please @jeremyjh take a look. I'd also be happy to write tests for this but for some reason tests are failing even if I go back in history before my previous PR, can you please take a look if it is failing for you? Maybe it's something with my setup.

### DIFF
--- a/test/fixtures/ignore_empty/mix.exs
+++ b/test/fixtures/ignore_empty/mix.exs
@@ -1,0 +1,15 @@
+defmodule Ignore.Mixfile do
+  use Mix.Project
+
+  def project do
+    [
+      app: :ignore,
+      version: "0.1.0",
+      prune_code_paths: false,
+      dialyzer: [
+        ignore_warnings: "ignore_empty_test.exs",
+        list_unused_filters: true
+      ]
+    ]
+  end
+end
--- a/lib/mix/tasks/dialyzer.ex
+++ b/lib/mix/tasks/dialyzer.ex
@@ -187,8 +187,14 @@ defmodule Mix.Tasks.Dialyzer do
           No :ignore_warnings opt specified in mix.exs. Using default: #{default}.
           """)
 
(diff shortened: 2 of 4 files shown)


jeremyjh:
@lnenad the tests do all pass for me locally. Can you paste a failure message? Yes I'd prefer to have tests and I think it should be possible following our project fixture patterns.

jeremyjh:
@lnenad  The issue with some of those is the line terminators are different on windows; its true the test suite isn't compatible with windows.  I'm not sure what the story is with format vs format_warning though; if that is a platform difference it is a surprising one. Any chance you can test in WSL or a VM with Linux?

jeremyjh:
@lnenad I think actually that those issues shouldn't prevent you from making a new test. If you can make a new test that passes for you in windows and fails with the old code, I can handle any issues with Linux compatibility/CI.

test/fixtures/ignore_empty/mix.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Ignore.Mixfile do
```
riverrun, line 1: Not sure, but there might be an issue with using the same module name `Ignore.Mixfile`. Maybe you can try changing one of these names, and see if that fixes some of these warnings.

pdm-jd:
I was linked to this PR after posting mine here https://github.com/jeremyjh/dialyxir/pull/549

It looks like a similar approach, but I'm not sure how to fix the failures in this one.

PR #550, opened 2024-11-13
update ex_doc to latest and benefits from better UI

epinault:
Exdoc is way behind. Newer UI is much better for searching and using.


epinault:
@jeremyjh can you approve? Also how do you release new version ? is this manually done?

Issue #551, opened 2024-12-18
Incorrect error reporting for invalid typespec


lud-wj:
Given [this example](https://hexdocs.pm/dialyxir/Dialyxir.Warnings.InvalidContract.html):

```elixir
defmodule Example do
  @spec process(:error) :: :ok
  def process(:ok) do
    :ok
  end
end
```

Dialyzer outputs the following:

```
lib/..../some-file.ex:14: Invalid type specification for function 'Elixir.WttjAi.IdealCandidateChatbot.Example':process/1.
 The success typing is 'Elixir.WttjAi.IdealCandidateChatbot.Example':process('ok') -> 'ok'
 But the spec is 'Elixir.WttjAi.IdealCandidateChatbot.Example':process('error') -> 'ok'
 They do not overlap in the 1st argument
```

But dialyxir outputs this:

```
lib/wttj_ai/ideal_candidate_chatbot.ex:14:invalid_contract
The @spec for the function does not match the success typing of the function.

Function:
WttjAi.IdealCandidateChatbot.Example.process/1

Success typing:
@spec process(:error) :: :ok
```

Success typing for dialyxir should be something like `process(:ok) :: :ok` instead of displaying the incorrect spec.

lud-wj:
@jimmybot this is fixed am I right?

jimmybot:
@lud-wj Yep, fixed!

lud-wj:
Thank you :)

PR #555, opened 2025-04-07
Fix invalid_contract to show success typing, not just repeating the spec

jimmybot:
For this error, the current output is simply the spec, which does not help by itself since the underlying error is that the success typing and the spec do not match. This modifies the warning message so that it shows both the success typing and the spec so that they can be compared and acted upon.

Fixes issue 551 reported here: https://github.com/jeremyjh/dialyxir/issues/551


jeremyjh:
Thanks!

PR #557, opened 2025-05-22
Specify more recent ex_doc

srcrip:
https://github.com/jeremyjh/dialyxir/pull/550 tried to update this as well but only the lock file was changed. I think the actual `mix.exs` needs to specify something greater than 'any version'.


jeremyjh:
Sorry I've neglected this so long, it was addressed in https://github.com/jeremyjh/dialyxir/pull/601 .

PR #558, opened 2025-05-28
Run with no formatter set

IslandUsurper:
When no formatter is specified, at least on OTP 26 and Elixir 1.16, calling `mix dialyzer` throws the following error

```
** (Protocol.UndefinedError) protocol Enumerable not implemented for nil of type Atom. This protocol is implemented for the following type(s): Date.Range, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, List, Map, MapSet, Range, Stream
    (elixir 1.16.3) lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir 1.16.3) lib/enum.ex:166: Enumerable.reduce/3
    (elixir 1.16.3) lib/enum.ex:4396: Enum.map/2
    (dialyxir 1.4.3) lib/dialyxir/dialyzer.ex:35: Dialyxir.Dialyzer.Runner.run/2
    (dialyxir 1.4.3) lib/dialyxir/dialyzer.ex:96: Dialyxir.Dialyzer.dialyze/3
    (dialyxir 1.4.3) lib/mix/tasks/dialyzer.ex:274: Mix.Tasks.Dialyzer.run_dialyzer/2
    (dialyxir 1.4.3) lib/mix/tasks/dialyzer.ex:205: Mix.Tasks.Dialyzer.run/1
    (mix 1.16.3) lib/mix/task.ex:478: anonymous fn/3 in Mix.Task.run_task/5
```

Accessing maps with `[]` returns `nil` when the key isn't present, instead of the empty list, so this change should work for all supported versions.


IslandUsurper:
Well, I finally figured out why it seemed I was the only one with this problem. At some point I had done a `mix archive.install dialyxir` and lately my project updated dialyxir in its `deps()` and the versions between them didn't match any more. So the way the configs were passed between the local version and the $MIX_HOME version didn't match up and it failed to set the default formatter.

Or something like that. Once I deleted the dialyxir in $MIX_HOME, it worked fine.

PR #560, opened 2025-06-11
Adds note about `:unknown` flag to `Default Flags` section from README

antedeguemon:
Hey there! 🖖 

I noticed a small issue when reading the [Flags section](https://github.com/jeremyjh/dialyxir?tab=readme-ov-file#flags) from the README:
> ### Flags
>
> _(..)_ As of 0.4, there are no longer any flags used by default. 

But the section below (and also Dialyxir's behavior) contradicts it:
>#### Dialyxir defaults
>
> By default `dialyxir` has always included the `:unknown` warning option so that warnings about unknown functions are returned. _(..)_

So I made a small wording update to clarify that 0.4 removed all flags except for `:unknown`, which is still included by underlying dialyxir's defaults.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,13 @@ Dialyxir supports formatting the errors in several different ways:
 
 ### Flags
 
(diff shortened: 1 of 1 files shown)


jeremyjh (approved):

jeremyjh:
Thanks!

PR #570, opened 2025-07-30
Fix failing CI runs

pwcsquared:
I noticed that the old version of the Ubuntu runner in the github workflow had been removed, so all CI runs are now hanging on the first step. I updated the runner to Ubuntu 24.04, added Elixir 1.18.4, and removed versions of Elixir and Erlang that had reached EOL and no longer receive security support from the testing matrix as well.

I have not added OTP 28 here since I suspect it will fail before https://github.com/jeremyjh/dialyxir/pull/569 merges, but once that branch is merged in OTP 28 should be good to add to the testing matrix as well.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
   check_duplicate_runs:
     name: Check for duplicate runs
     continue-on-error: true
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     outputs:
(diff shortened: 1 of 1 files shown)


pwcsquared:
@jeremyjh There are some pending checks for older Elixir versions that I removed in this PR. If removing those versions is acceptable, you may need to adjust the branch protection rules on the repo

jeremyjh:
Thanks!

Issue #223, opened 2018-07-20
Deprecate :project and :transitive dependencies


jeremyjh:
These really only exist for backwards compatibility. Not sure if they should just be removed from docs or get a deprecation warning, but at least one new user on the forums was using `:transitive` so leaning towards a warning.

asummers:
We use transitive in the code at work. Does it not actually do anything?

jeremyjh:
Well what it does is include your mix dependency tree instead of your OTP dependency tree. This can easily include things you don't need, like dev tools.

jeremyjh:
Yes that is how it works, but the default `app_tree` does the same thing but using OTP application trees. This is more inclusive in the sense it will also bring in OTP applications that are not specified in mix (such as `:mnesia`), and less inclusive in that it leaves out anything with `runtime: false`.  If people have OTP dependencies that are missing or not properly specified their deployment fails when using any OTP release package, so it could help you catch that earlier.

asummers:
Interesting. I have no a priori reason to say no, then, I don't think.

mguimas:
Well the docs say `:transitive - Include Mix and OTP application dependencies recursively`, so I am using `:transitive` to have Mix and its deps, and these deps' deps, and so on, and also my application deps tree.

If `:transitive` goes away how can I get the above effect?

PR #575, opened 2025-10-29
feat: add incremental mode support for Dialyzer (OTP 26+)

### LINKED ISSUE #498: OTP 26: Incremental mode for Dialyzer: --incremental
[OTP 26 was released today, along with the new `--incremental` flag for Dialyzer](https://www.erlang.org/blog/otp-26-highlights/#incremental-mode-for-dialyzer).

This was first discussed in https://github.com/jeremyjh/dialyxir/issues/105. Now that OTP 26 is out, I figured I'd make a new issue to track this.

If I find some time I'll start a branch to test out the new `--incremental` mode within `dialyxir`.

As for usage/API, I think we can:
- [Add a `--incremental` CLI option](https://github.com/jeremyjh/dialyxir#command-line-options).
- Add either an `incremental: true` or `mode: :incremental` flag for `mix.exs` config.

Ch4s3:
### Summary

This PR introduces support for **Dialyzer’s [incremental mode](https://github.com/erlang/otp/commit/963c7d5ccecb9ea4940e7112d88534781a413fa6)** (OTP 26+) in Dialyxir, including the ability to run incremental analysis based on passed in OTP applications.

The goal is to make Dialyxir work naturally with Dialyzer’s newer incremental workflow while remaining backwards compatible with existing setups.

---

### What is this for

Dialyzer’s incremental mode is designed around:

* long-lived, incremental PLTs
* app-based analysis using `--apps` / `--warning_apps`
* dependency-aware re-analysis

Dialyxir historically drives Dialyzer using classic PLTs and a whole app analysis. Incremental mode can significantly improve analysis of smaller changes on large code bases especially in CI environments.

---

#### 1. Incremental mode integration

Dialyxir can now pass `incremental: true` through to Dialyzer, enabling OTP 26+ incremental analysis.

#### 2. Application-based analysis support

New support for Dialyzer’s:

* `--apps`
* `--warning_apps`

via Dialyxir configuration.

#### 3. Higher-level configuration options

New symbolic configuration values:

```elixir
dialyzer: [
  incremental: true,
  core_apps: [:erts, :kernel, :stdlib, :crypto, :public_key, :ssl, :elixir, :logger, :mix],
  apps: :transitive,
  warning_apps: :project
]
```

Supported behaviors:

* `apps: :project` → current project / umbrella apps
* `apps: :transitive` → `core_apps ++ deps ++ project_apps`
* `warning_apps: :project` → only project apps
* `warning_apps: :all` → all resolved apps

This allows large projects to opt into app-based incremental mode without manually maintaining long, fragile app lists.

---

### Implementation overview

* Adds app-resolution logic in `Dialyxir.Project` to compute:

  * project apps (umbrella or single)
  * runtime dependency apps (via public Mix project config)
  * user-defined core apps (mostly things from OTP plus elixir itself)

* Introduces resolution helpers:

  * `resolve_apps/1`
  * `resolve_warning_apps/2`
  * `resolve_app_args/2`

* Updates the Mix task to:

  * resolve symbolic config into concrete app lists
  * automatically switch Dialyzer into app-mode when appropriate
  * drop `:files` when running app-mode to avoid mixed invocation types.

---

### Backwards compatibility

This change is opt-in:

* If incremental mode is not flagged/configured, Dialyxir behaves exactly as before.
* Existing plt based Dialyzer workflows are unaffected.
* App-based incremental mode only activates when the new config/flag values are used.

---

### How to use

Example umbrella setup:

```elixir
dialyzer: [
  incremental: true,
  core_apps: [:erts, :kernel, :stdlib, :crypto, :public_key, :ssl, :elixir, :logger],
  apps: :transitive,
  warning_apps: :project
]
```

Run:

```bash
mix dialyzer --incremental
```

This enables Dialyzer’s native incremental mode

#### Large codebase initial run
```

our_app
  Dialyzer exited with code 2
  Incremental mode enabled; skipping PLT check step
  Will use PLT file: ops/plts/dialyxir_erlang-28.1.1_elixir-1.19.3_deps-test_incremental.plt
  ignore_warnings: .dialyzer_ignore.exs
  Starting Dialyzer
  [
    analysis_type: :incremental,
    warning_apps: [:monitoring, :api, :another_app, :another_app_2,
     :another_app_3, :another_app_4, ...],
    ...
  ]
  Total errors: 68, Skipped: 65, Unnecessary Skips: 5
  done in 25m3.7s
  Warning: The created anonymous function has no local return.
  Warning: The function call update! will not succeed.
  Warning: Invalid type specification for function changeset.
  done (warnings were emitted)
  Halting VM with exit status 2
Error: Process completed with exit code 1.
```

#### Large app 2nd run with error fix

```

our_app
  Dialyzer exited with code 0
  Incremental mode enabled; skipping PLT check step
  Will use PLT file: ops/plts/dialyxir_erlang-28.1.1_elixir-1.19.3_deps-test_incremental.plt
  ignore_warnings: .dialyzer_ignore.exs
  Starting Dialyzer
  [
    analysis_type: :incremental,
    warning_apps: [:monitoring, :api, :another_app, :another_app_2,
     :another_app_3, :another_app_4, ...],
    ...
  ]
  Total errors: 65, Skipped: 65, Unnecessary Skips: 5
  done in 0m23.13s
  ```

**You can see that the second run with a `+1/-0` diff ran in 23.13s.** This application has around 700k lines of elixir code.

### Demonstration App

[github](https://github.com/Ch4s3/test_umbrella_app)
[ci run](https://github.com/Ch4s3/test_umbrella_app/actions/runs/19942860739/job/57187559511)

Refs: https://www.erlang.org/doc/apps/dialyzer/dialyzer.html#incremental-mode
Closes: #498

### CODE REVIEW

**Comment by Ch4s3** [comment]
I'm still working on the tests,

**Comment by oliver-kriska** [comment]
I think adding good documentation is base. Right now I don't see how library can give some tool or approach for it. Because it depends on own CI/CD configurations how people use cache. Something can be handled by library only in case library is able to set how to store these incremental files. Thanks library can partly allow to configure it for developers or use some common cache so developers will not have to configure new cache or something.

**Comment by jeremyjh** [comment]
> @jeremyjh do you have any pointers for getting the tests to not time out in CI? They all work for me locally.

@Ch4s3 Thanks for working on this!

The problem must be in this branch; I reran master - it passed and test times haven't changed. https://github.com/jeremyjh/dialyxir/pull/578

I don't immediately see the issue; there are tests timing out that aren't really doing much. I think I'd start by commenting all the new tests; if it passes then uncomment just the new output test.

**Comment by Ch4s3** [comment]
@oliver-kriska once I have tests working, I'm going to test this in CI on my app and gather timing info as well as double checking the path information and I'll update docs here.

File: test/mix/tasks/dialyzer_test.exs [hunk 1]
```diff
@@ -103,4 +103,49 @@ defmodule Mix.Tasks.DialyzerTest do
...
+  test "halts with error when incremental mode enabled on OTP < 26" do
+    # This test would only be meaningful on OTP < 26
+    # On OTP 26+, no error should be displayed
+    otp_version = :erlang.system_info(:otp_release) |> List.to_string() |> String.to_integer()
+
+    if otp_version < 26 do
+      in_project(:incremental, fn ->
+        # On OTP < 26, the process should halt with exit code 3
+        assert_raise SystemExit, fn ->
+          Mix.Tasks.Dialyzer.run(["--ignore-exit-status"])
```
Ch4s3:
```suggestion
          Mix.Tasks.Dialyzer.run(["--incremental"])
```

Ch4s3:
I think I found the main problem

Ch4s3:
I'm still working on this, I just had to take a break do to work/life business.

Ch4s3:
I think the issue here was trying to test the system halt call on older OTP versions. Since the other code path that calls a halt isn't tested, I'm removing the test that verifies a halt on OTP < 26. Hopefully that's an acceptable tradeoff.

Ch4s3:
I'm fiddling with how to correctly use this in CI. It seems like just `mix dialyzer --incremental` plus any formatting should be run without a discrete `--plt` step. It doesn't seem like OTP generates an incremental plt if you use both flags.

michalmuskala:
Yes, with the new incremental mode you don't need a separate PLT step - an incremental run is both generating a new PLT and checking your code - internally it was always the same operation, this just explicitly combines them.

Ch4s3:
@michalmuskala  thanks for weighing in. Am I understanding correctly that incremental mode doesn't build the core language PLT so you need to also have that if you want to avoid getting errors from missing language functions?

Or should we be using `--apps` and `--warning-apps` to deal with this?

TD5:
> Am I understanding correctly that incremental mode doesn't build the core language PLT so you need to also have that if you want to avoid getting errors from missing language functions?
> Or should we be using --apps and --warning-apps to deal with this?

Incremental mode analyses whatever modules you give it, including OTP modules (they don't get any special treatment). That means that you do want to use `--apps` and `--warning-apps` to make sure everything is included. For code that you don't own (OTP, libraries made by others), I'd suggest using `--apps`, so Dialyzer knows about them and can comment on their usage, but won't complain about them directly. Then use `--warning-apps` for code you actually want issues to be raised for.

Sadly, it's not quite as clean and orthogonal as it sounds, because if there is some sort of discrepancy in your `--apps`, you won't get a direct error reported for it, but your usages of that code from your `--warning-apps` might end up generating warnings due to your calls into broken code. As such, it's worth having that in the back of your mind when investigating Dialyzer warnings.

For the mental model for how to use `--incremental` mode, try to forget your existing knowledge of how Dialyzer works, and instead just list all the code you care about as either `--apps` or `--warnings-apps` (including OTP and libraries). You call the same command each time, asking Dialyzer to analyse your`--apps` and `--warnings-apps`. The incremental command is given a path to a PLT file, and it will worry about building, changing and updating the cache (PLT) file to minimise the work it needs to use between runs. You don't need to do anything explicitly to manage that lifecycle apart from making sure you point it to a consistent PLT file (and as someone pointed out above, perhaps you want to point to a different one per build configuration/branch if you're often moving between them).

README.md
```diff
@@ -212,6 +213,101 @@ def project do
 end
 ```
 
+### Incremental Mode
+
+Dialyxir supports Dialyzer's incremental analysis mode (available in OTP 26+). When enabled, Dialyzer will reuse previous analysis results and only analyze changed modules, significantly speeding up subsequent runs.
```
TD5: Strictly, in incremental mode, Dialyzer analyses the modules that were changed since the PLT was last used (if there is one), _plus_ the set of modules which Dialyzer thinks might depend on it. This perhaps sounds like a trivial difference, but if you have a core module that is used by a lot of your codebase, changing that will cause all the modules that depend on it to need to be re-analysed too (since you might have fixed/broken them with your change!).

In my experience, this leads to people asking: "I changed this one file but Dialyzer analyses hundreds of files - wasn't incrementality supposed to fix this?", with the answer being that incrementally tries to reduce re-analysing _irrelevant_ files, but files can be relevant because they're in a dependency chain with something that has changed.

README.md
```diff
@@ -212,6 +213,101 @@ def project do
...
+  incremental: true,
+  plt_incremental_file: "_build/dev/my_incremental.plt"
+]
+```
+
+If `plt_incremental_file` is not specified, Dialyxir automatically appends `_incremental` to the base PLT filename.
+
+**Analyzing specific applications:**
+
+When using incremental mode, you can specify which applications to analyze using the `apps` and `warning_apps` options. This allows you to analyze entire applications, which can be more efficient for large codebases.
```
TD5: > This allows you to analyze entire applications, which can be more efficient for large codebases.

Strictly, this is true, but can lead to misleading results if you include modules to be analysed with the modules they (transtively) use function or type definitions from. My suggestion would be to _analyse all your code in a repo, including any dependencies it has, in one go_. Incrementality itself will do the heavy lifting to soundly work out what code isn't affected by a change, avoiding the risk of the mistake above. Of course, at a certain scale you may really have no choice (e.g. if your codebase is so large the Dialyzer analysis for the entire codebase becomes too big to fit into RAM).

README.md
```diff
@@ -212,6 +213,101 @@ def project do
...
+
+**How `apps` and `warning_apps` work:**
+
+- `apps` specifies which applications to analyze. All applications listed here will be analyzed.
+- `warning_apps` (optional) controls which applications will emit warnings:
+  - If `warning_apps` is **not** specified, warnings are reported for all applications in `apps`.
+  - If `warning_apps` **is** specified, only those applications will have warnings reported.
+  - Applications in `apps` but not in `warning_apps` are still analyzed (to provide context for the analysis), but warnings will not be reported for them.
+
+This is useful when you want to include dependencies in the analysis (so Dialyzer can find discrepancies in how you use them), but only see warnings for your own code.
```
TD5: In my mind, this is the primary motivation for `apps` and `warning_apps`, rather than trying to only analyse a subset of a codebase (since that's so error prone).

TD5:
Hi there 👋 I made Dialyzer's incremental mode. I tried to add a bit of context where I could on this PR. I hope that helps.

README.md
Ch4s3: > This perhaps sounds like a trivial difference

No, this makes perfect sense. I was struggling with how to explain this well, perhaps even to myself.

> but files can be relevant because they're in a dependency chain with something that has changed

Right, the classic problem that changing a config file causes a massive recompilation.

Ch4s3:
> > Am I understanding correctly that incremental mode doesn't build the core language PLT so you need to also have that if you want to avoid getting errors from missing language functions?
> > Or should we be using --apps and --warning-apps to deal with this?
> 
> Incremental mode analyses whatever modules you give it, including OTP modules (they don't get any special treatment). That means that you do want to use `--apps` and `--warning-apps` to make sure everything is included. For code that you don't own (OTP, libraries made by others), I'd suggest using `--apps`, so Dialyzer knows about them and can comment on their usage, but won't complain about them directly. Then use `--warning-apps` for code you actually want issues to be raised for.

This is what I'm figured out after some trial and error and the latest changes are an attempt to make passing those through as straightforward as possible. Thanks for further clarification.


> Sadly, it's not quite as clean and orthogonal as it sounds, because if there is some sort of discrepancy in your `--apps`, you won't get a direct error reported for it, but your usages of that code from your `--warning-apps` might end up generating warnings due to your calls into broken code. As such, it's worth having that in the back of your mind when investigating Dialyzer warnings.

Makes sense.

> For the mental model for how to use `--incremental` mode, try to forget your existing knowledge of how Dialyzer works,  and instead just list all the code you care about as either `--apps` or `--warnings-apps` (including OTP and libraries). You call the same command each time, asking Dialyzer to analyse your`--apps` and `--warnings-apps`. The incremental command is given a path to a PLT file, and it will worry about building, changing and updating the cache (PLT) file to minimise the work it needs to use between runs. You don't need to do anything explicitly to manage that lifecycle apart from making sure you point it to a consistent PLT file (and as someone pointed out above, perhaps you want to point to a different one per build configuration/branch if you're often moving between them).

Awesome, this is such helpful context! I think I'm pretty close on this, I have the branch working against a production repo and giving me meaningful output. Thanks again, and also for your work on incremental mode.

README.md
Ch4s3: these are great comments, I think I can do a lot better with the readme now. thanks!

Ch4s3:
@TD5 could you weigh in again? I'm now just passing --apps and --warning-apps but not the files in incremental mode and I'm not seeing errors for code with trivial dialyzer errors picked up by classic mode.

Based on [this test setup](https://github.com/erlang/otp/blob/4ece5771e493186dc88b4641090229a6cabd2a37/lib/dialyzer/test/incremental_SUITE.erl#L1038) from OTP, and the [original commit message](https://github.com/erlang/otp/commit/963c7d5ccecb9ea4940e7112d88534781a413fa6)  I'm assuming --apps and --warning-apps have to overlap. I'm not sure how I missed this before.

TD5:
> Based on [this test setup](https://github.com/erlang/otp/blob/4ece5771e493186dc88b4641090229a6cabd2a37/lib/dialyzer/test/incremental_SUITE.erl#L1038) from OTP, and the [original commit message](https://github.com/erlang/otp/commit/963c7d5ccecb9ea4940e7112d88534781a413fa6) I'm assuming --apps and --warning-apps have to overlap. I'm not sure how I missed this before.

That sounds plausible. Honestly, I wrote the code long enough ago that I can't remember the motivation there.

Ch4s3:
> That sounds plausible. Honestly, I wrote the code long enough ago that I can't remember the motivation there.

I know the feeling! Thanks for the input.

lib/dialyxir/project.ex
```diff
@@ -114,6 +147,165 @@ defmodule Dialyxir.Project do
...
+  def dialyzer_warning_apps do
+    config = dialyzer_config()
+    resolved = resolve_warning_apps(config)
+
+    # Maintain backward compatibility: if config is already a list, return it as-is
+    # If resolved is nil, return empty list for backward compatibility
+    cond do
+      is_list(config[:warning_apps]) -> config[:warning_apps]
+      resolved == nil -> []
+      true -> resolved
```
Ch4s3: I'm not crazy about this and may need to change it.

lib/mix/tasks/dialyzer.ex
```diff
@@ -308,6 +458,88 @@ defmodule Mix.Tasks.Dialyzer do
...
+
+  defp maybe_put_warning_apps(opts, []), do: opts
+  defp maybe_put_warning_apps(opts, apps), do: Keyword.put(opts, :warning_apps, apps)
+
+  defp maybe_put_files(args, apps, _warning_apps) do
+    cond do
+      apps != [] ->
+        # Application-based mode: when apps are provided, do NOT pass files.
+        # --apps and --files are mutually exclusive modes in Dialyzer.
+        # Dialyzer will resolve app names to their BEAM files internally.
```
Ch4s3, line 468: I'm like 95% sure this is true, but I'm not great at erlang so I could have misread the code in OTP.

lib/mix/tasks/dialyzer.ex
```diff
@@ -308,6 +458,88 @@ defmodule Mix.Tasks.Dialyzer do
...
+      #{inspect(missing_apps)}
+
+      This may cause Dialyzer to miss type information from these applications.
+      """)
+    end
+
+    valid_apps
+  end
+
+  defp app_exists?(app) do
```
Ch4s3: this feels a bit defensive but I tripped myself up on this a few times so I added a check.

Ch4s3:
@jeremyjh this should basically work now and I like the api, with the caveat that core_apps feels a bit hacky. I'm putting it down until after US Thanksgiving, but it should be in a reviewable state even though I plan to refactor a bit next week.

lib/dialyxir/project.ex
```diff
@@ -114,6 +147,165 @@ defmodule Dialyxir.Project do
...
+    end
+  end
+
+  # Returns core apps configured under :core_apps.
+  # This is a Dialyxir-only concept used for incremental app mode.
```
jeremyjh: @Ch4s3 I haven't really had a chance to dig into the details here. Will the new `core_apps` key only be used for incremental mode or would it apply to normal builds as well? If we want a separate apps list for incremental maybe we should namespace the options so thats more clear.

Ch4s3: That's the idea, but I really don't love it. I can definitely namespace it, but I'm not 100% sure yet that its the right options api for this problem. I just couldn't find a reliable way to programatically get these.

jeremyjh: Why do we need a different app list for incremental mode? Could we not use whatever the current options API resolves ?

Ch4s3: we could use `:apps` but I wanted to support `apps: :transitive` and couln't figure out a good way to automatically pick up things like `:erts`, and `:elixir` but I'm afraid it's too clever by half. I'm exploring other options.

Ch4s3: without `:elixir` in apps you end up with missing function warnings for thing like `String.t()`.

Ch4s3: I have some new changes shaping up, and I'll publish a test app with this running in CI perhaps later this week.

lib/dialyxir/plt.ex
```diff
@@ -241,10 +241,15 @@ defmodule Dialyxir.Plt do
...
     case plt_info(plt) do
-      {:ok, info} ->
+      {:ok, info} when is_list(info) ->
         Keyword.fetch!(info, :files)
         |> Enum.reduce(MapSet.new(), &MapSet.put(&2, Path.expand(&1)))
 
+      {:ok, {:incremental, _modules}} ->
+        # Incremental PLTs are managed by Dialyzer itself and have a different structure
+        # Return nil to indicate we can't enumerate files from incremental PLTs
+        nil
```
Ch4s3: this can be supported, I just haven't tried yet

lib/mix/tasks/dialyzer.ex
Ch4s3: This is correct.

Ch4s3:
@jeremyjh I have a coworker sanity checking the code, and I'm doing a little doc cleanup and refactoring, but we've confirmed this works locally on our macs, in ci for our large app, and in my test repo on GHA.

notactuallytreyanastasio:
We ran into a REALLY annoying issue that was coming from default low ulimits on Mac OS. I made a PR to @Ch4s3's branch that, hopefully, will make it so no one else loses hours on this wondering why it passes on CI but not on some random Mac lol 

https://github.com/Ch4s3/dialyxir/pull/1

Ch4s3:
> We ran into a REALLY annoying issue that was coming from default low ulimits on Mac OS. I made a PR to @Ch4s3's branch that, hopefully, will make it so no one else loses hours on this wondering why it passes on CI but not on some random Mac lol
> 
> [Ch4s3#1](https://github.com/Ch4s3/dialyxir/pull/1)

I think this may be best to handle in the readme rather than in code.

Ch4s3:
@jeremyjh I think this is reviewable now. Below is a sketch of how to read through the changes or as best I could reason about how to approach it.

### 1) Start with the intent and surface changes
- **Changelog**: `CHANGELOG.md` — confirms incremental mode feature addition and headline behavior.
- **README incremental section**: `README.md` — updated guidance for `apps`/`warning_apps`, incremental usage, caching, and macOS ulimit tip.

### 2) Core entry points (behavioral changes)
- **Mix task front door**: [`lib/mix/tasks/dialyzer.ex`](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-b5ee974e03ac9fe59b1c6db3ac5b458186ca5096d0127f226fc3e181bcefa633R226)
  - CLI parsing and incremental gating: [R226](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-b5ee974e03ac9fe59b1c6db3ac5b458186ca5096d0127f226fc3e181bcefa633R226).
  - Selection assembly and resolver wiring: [R252](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-b5ee974e03ac9fe59b1c6db3ac5b458186ca5096d0127f226fc3e181bcefa633R252).
  - Argument assembly for Dialyzer: [R319](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-b5ee974e03ac9fe59b1c6db3ac5b458186ca5096d0127f226fc3e181bcefa633R319) and `dialyzer_args/4` around [R488](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-b5ee974e03ac9fe59b1c6db3ac5b458186ca5096d0127f226fc3e181bcefa633R488).
- **Dialyzer runner**: [`lib/dialyxir/dialyzer.ex`](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-33a35df1e4c5545b1bc1f078a4e4952fe6be18a5e19152d2bac167d9048740c7R12)
  - Incremental analysis toggle and formatter parsing: [R12](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-33a35df1e4c5545b1bc1f078a4e4952fe6be18a5e19152d2bac167d9048740c7R12).

### 3) Centralized app/warning_app resolution
- **New resolver module**: `lib/dialyxir/app_selection.ex`
  - Single place that merges CLI + config, expands flags, normalizes atom/string/charlist, filters non-project `warning_apps` with warnings, and merges `warning_apps` into `apps`.
  - Side-effect scope: only emits warnings via `Dialyxir.Output`.
- **Project-level helpers**: [`lib/dialyxir/project.ex`](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-cca149b77f9bf10c032ddc175e03118da0928b88c71ffb338db43f6dcbe6b8ecR307)
  - `resolve_apps/1` and `resolve_warning_apps/1` unify config resolution; see [R307](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-cca149b77f9bf10c032ddc175e03118da0928b88c71ffb338db43f6dcbe6b8ecR307) and [R337](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-cca149b77f9bf10c032ddc175e03118da0928b88c71ffb338db43f6dcbe6b8ecR337).
  - `fallback_list/3` and `resolve_list_value/2` handle list configs and backward compatibility; see [R354](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-cca149b77f9bf10c032ddc175e03118da0928b88c71ffb338db43f6dcbe6b8ecR354).
  - `expand_transitive_apps/1` extracts `:transitive` expansion to avoid duplication; see [R375](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-cca149b77f9bf10c032ddc175e03118da0928b88c71ffb338db43f6dcbe6b8ecR375).
  - Semantics: `apps` expands `:transitive` (deps + project) but still requires explicit OTP apps; `warning_apps` lists are project-only, merged into `apps` later; nil → empty list for backward compatibility.

### 4) Incremental PLT handling
- **PLT file handling**: [`lib/dialyxir/plt.ex`](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-cca6492d9081491c7f18af35cb5ae01da5a90a23b38faaafbc52924abafb28c8R244)
  - Confirm file enumeration and incremental PLT handling are compatible with incremental analysis (no regressions in classic PLTs): see [R244](https://github.com/jeremyjh/dialyxir/pull/575/files#diff-cca6492d9081491c7f18af35cb5ae01da5a90a23b38faaafbc52924abafb28c8R244) for classic vs incremental PLT file handling.

### 5) Docs and CI caching
- **CI examples**: `docs/github_actions.md`, `docs/gitlab_ci.md`, `docs/circleci.md`
  - Clarified cache keys (OTP/Elixir/mix.lock, optional branch keys) and guidance that incremental artifacts live in `priv/plts`.
- **MacOS ulimit note**: `README.md`
  - `EMFILE` avoidance tip for large codebases during MD5 hashing.

### 6) Tests (behavioural coverage)
- **Resolver unit tests**: `test/dialyxir/app_selection_test.exs`
  - Covers incremental gating, CLI overrides, filtering, and merging.
- **Project helpers**: `test/dialyxir/project_test.exs`
  - Expanded scenarios for `apps`/`warning_apps` resolution and `:transitive` expectations.
- **Task integration**: `test/mix/tasks/dialyzer_test.exs`
  - Ensures flags flow into dialyzer args correctly, including incremental/app-mode paths.
- **Dialyzer runner**: `test/dialyxir/dialyzer_test.exs`
  - Confirms runner behavior in incremental contexts.

### 7) Fixtures and samples
- **Incremental/app-mode fixtures**: under `test/fixtures/*` (e.g., `apps_transitive`, `warning_apps_project`, `incremental_*`)
  - Check they mirror the intended semantics (OTP apps explicit, `warning_apps` limited to project).

### 8) Thing to look out for
- CLI vs config precedence in `AppSelection`: CLI should win; `warning_apps` filtered to project apps only; `apps` expanded for `:transitive`.
- Incremental vs classic: incremental should skip classic PLT build when `--incremental`; classic flow unchanged when not incremental.
- Caching guidance: `priv/plts` is the cache root for both classic and incremental artifacts; keys tied to OTP/Elixir/mix.lock (optional branch suffix).
- ulimit/EMFILE: documented; no code changes, just awareness for large repos.

test/fixtures/core_apps_config/mix.exs
```diff
@@ -0,0 +1,15 @@
...
+  use Mix.Project
+
+  def project do
+    [
+      app: :core_apps_config,
+      version: "0.1.0",
+      prune_code_paths: false,
+      dialyzer: [
+        incremental: true,
+        core_apps: [:erts, :kernel, :stdlib, :crypto, :public_key, :ssl, :elixir, :logger]
```
ypconstante: this config was removed

README.md
```diff
@@ -212,6 +213,166 @@ def project do
...
+> **Note:** if your `apps` list is incomplete or inconsistent, you won't get a
+> direct error for that – you may just see confusing warnings in your
+> `warning_apps` because Dialyzer inferred "broken" behaviour for the missing
+> modules. Keep this in mind when debugging surprising incremental warnings.
+
+**Configuration options:**
+
+Both `apps` and `warning_apps` accept:
+- An explicit list of apps: `[:app1, :app2, ...]`
+- The `:transitive` flag – automatically includes all dependencies + project apps
```
jeremyjh: Both :transitive and :project modes are deprecated in favor of `:app_tree` (the default, equivalent to `mix app.tree`) and `:apps_direct` (direct app dependencies). In most cases we only want to include only apps that would be deployed in the application, and not other project dependencies.

https://github.com/jeremyjh/dialyxir/issues/223

README.md
```diff
@@ -212,6 +213,166 @@ def project do
...
+- cache the incremental PLT path per OS + OTP + Elixir + lockfile, and
+- optionally include the branch name in the cache key, so long-lived feature
+  branches can benefit from warm incremental PLTs without polluting `main`.
+
+**Analyzing specific applications**
+
+When using incremental mode, you can tell Dialyzer which applications to analyse
+using the `apps` and `warning_apps` options:
+
+- `apps` – all applications that Dialyzer should know about and analyse upstream of your own code
```
jeremyjh, line 316: If this key is only meaningful for incremental mode we need to indicate that in the option name, or nest it e.g. `incremental: [apps: [...]]`. 

Where possible - where the semantics are the same - we should use the same config names as https://github.com/jeremyjh/dialyxir?tab=readme-ov-file#dependencies - and let them optionally nested to override incremental mode. This will make adoption easier for people.

jeremyjh (changes requested):

README.md
Ch4s3: not sure how I missed this, I'll get it addressed

README.md
Ch4s3: I think I'll try nesting it. I looked at using plt_add_deps and plt_add_app but the semantics were a bit different and it made the existing code really hard to follow.

README.md
Ch4s3: I think in most cases we want to add `some_otp_stuff ++ [:app_tree]` to `:apps`

Ch4s3: my one concern is that `:apps_direct` includes deps and for the way `:warning_apps` is used in incremental mode you need just a list of your own apps and not the deps.

I really need to either have another flag here or to just require users to declare their app/apps. It's mostly only an issue for umbrella apps though.

Ch4s3: I've added a new key, `:apps_project` that uses `Mix.Project.apps_paths()` (if defined) to get top level project applications. It only works in `:warning_apps` and is documented as such.

Ch4s3:
@jeremyjh there are some sort of tricky tradeoffs to consider here. The existing code for resolving dependencies is built around the way classic PLTs work and assumes that things like `:erts`, `:elixir`, `:kernel`, and so on are already in the PLT. This means that trying to add those items in automatically through the old code causes runtime warnings about loading deps like the following:

```
Error loading sasl, dependency list may be incomplete.
 {~c"no such file or directory", ~c"sasl.app"}
 ```
 
 I can solve that by resolving deps in a less efficient way in a new code path, or by requiring the end user to declare them explicitly in `apps: [...]`. Any other approach would have to cut into a lot of older code in a way that feels dicey in terms of broad ecosystem support (to me at least).
 
 
 ### Update
 
 I think this is ready for re-review. My compromise here is that users have to declare core language dependencies like `[:erts, :kernel, :stdlib, :elixir, :logger]`. I don't love it but trying to make it work automatically with existing dialixyr code using an existing flag would require a lot of changes to dependency resolution. You can see it in use [here](https://github.com/Ch4s3/test_umbrella_app/blob/76236d5f72a257cafd734a53491b96c465fd4d45/mix.exs#L16)

**Comment by jeremyjh** [comment]
@Ch4s3 is there an issue with always adding `[:erts, :kernel, :stdlib, :crypto, :elixir]` to the app list when `:app_tree` or `:apps_direct` are used? If a user doesn't want some of those, they could just provide a complete list of the apps they want instead of using either of those methods. I'm not sure that is an edge case that even exists.

**Comment by Ch4s3** [comment]
> @Ch4s3 is there an issue with always adding `[:erts, :kernel, :stdlib, :crypto, :elixir]` to the app list when `:app_tree` or `:apps_direct` are used? If a user doesn't want some of those, they could just provide a complete list of the apps they want instead of using either of those methods. I'm not sure that is an edge case that even exists.

@jeremyjh if the hard coded value is good with you I can absolutely add it to app_tree. I was just trying to avoid hard coding if I could.

**Comment by jeremyjh** [comment]
> @jeremyjh if the hard coded value is good with you I can absolutely add it to app_tree. I was just trying to avoid hard coding if I could.

It is already hard coded, those are the default apps that go into the Erlang/Elixir PLTs and are included in the project one.

**Comment by Ch4s3** [comment]
@jeremyjh I assumed since it was deprecated that I should do something different. I'll add it into my code path.

If you're  you ok using the new apps_project flag for the warning apps then it's read for review.

**Comment by Wigny** [comment]
Hey folks, how is this work going? I'm excited to have this ready to use and try it out! Is there something we can do to make that happen?

**Comment by Ch4s3** [comment]
@Wigny I'm just waiting for now. I've released an incremental only dialyzer wrapper at https://github.com/Ch4s3/assay that's experimental if you want to try that while you wait for this PR to be merged.

**Comment by ypconstante** [comment]
I did some testing in a project I'm working on, and got some issues with the new configs.

When I tried to use `:app_tree` I got an error because of sasl:
```
:dialyzer.run error: No such file, directory or application: "sasl"
```

When I tried to use `:apps_direct` I got an error because of credo:
```
:dialyzer.run error: No such file, directory or application: "credo"
```

Dialyzer is running with prod env, and credo is available only on test env.

**Comment by Ch4s3** [comment]
> I did some testing in a project I'm working on, and got some issues with the new configs.
> 
> When I tried to use `:app_tree` I got an error because of sasl:
> 
> ```
> :dialyzer.run error: No such file, directory or application: "sasl"
> ```
> 
> When I tried to use `:apps_direct` I got an error because of credo:
> 
> ```
> :dialyzer.run error: No such file, directory or application: "credo"
> ```
> 
> Dialyzer is running with prod env, and credo is available only on test env.

Interesting, I must not have considered running this with `MIX_ENV=prod`. Is that how you were running dialyzer before?

**Comment by ypconstante** [comment]
Yeah, in this project we only run dialyzer with `MIX_ENV=prod`, and it works with just this setting `dialyzer: [plt_add_apps: [:mix, :phoenix_storybook]]`

**Comment by Ch4s3** [comment]
> Yeah, in this project we only run dialyzer with `MIX_ENV=prod`, and it works with just this setting `dialyzer: [plt_add_apps: [:mix, :phoenix_storybook]]`

Gotcha, I hadn't considered running in prod, I'm a little surprised it ever worked. This is [the line that causes the issue](https://github.com/jeremyjh/dialyxir/pull/575/changes#diff-5c19459efa80671769235504a19c2db6bf014ec50b0b644fc6c1ee084409592bR349). I'm not sure there's a way to meet @jeremyjh 's desire to make the helpers "just work" and to have incremental mode work in `MIX_ENV=prod`. Incremental mode in dialyzer doesn't have plt_apps, and I use these helpers like `app_tree` to sort of fake it.

If you can suggest a clean fix, I'm open.

**Comment by oliver-kriska** [comment]
### `plt_ignore_apps` not respected in incremental mode

I tested this PR branch on our production codebase (Elixir 1.19.5, OTP 28) and incremental mode works — great work! The speedup is significant:

| Run | Mode | Time |
|-----|------|------|
| Classic (existing PLT) | Classic | **5m 34s** |
| Incremental 1st run | Incremental | **57m 15s** (initial PLT build) |
| Incremental 2nd run (no changes) | Incremental | **21s** |
| Incremental 3rd run (1 file changed) | Incremental | **20s** |

However, I found an issue with `plt_ignore_apps` being silently ignored in incremental mode.

**Our config:**
```elixir
dialyzer: [
  plt_ignore_apps: [:gpt3_tokenizer, :memoize],
  # ...
  incremental: [
    enabled: true,
    apps: [:elixir, :logger, :crypto, :public_key, :ex_unit, :credo, :tidewave] ++ [:app_tree],
    warning_apps: [:enaia]
  ]
]
```

**What happens:**

In classic mode, `plt_ignore_apps` excludes `gpt3_tokenizer` from the PLT (it caused an infinite loop during classic PLT building). Since dialyzer doesn't know about the module, calls to `Gpt3Tokenizer.token_count/1` produce a "function does not exist" warning, which we suppress via a regex filter in our ignore file.

In incremental mode with `:app_tree`, the new `AppSelection` module resolves apps independently and **does not subtract `plt_ignore_apps`**. So `gpt3_tokenizer` gets included, dialyzer sees the function exists, no warning is generated, and our ignore filter becomes **unused** — which causes `list_unused_filters: true` to fail CI with exit code 1.

Looking at the code, `lib/dialyxir/project.ex:91` applies `plt_ignore_apps()` in the classic path via `cons_apps/0`, but `lib/dialyxir/app_selection.ex` has no reference to `plt_ignore_apps` at all.

**Suggested fix:**

`AppSelection` should read `plt_ignore_apps` from config and subtract them from the resolved app list, consistent with the classic path. This would make migration from classic → incremental seamless for users who rely on `plt_ignore_apps`.

In our case the infinite loop turned out to be PLT-specific and didn't occur in incremental mode, but the behavioral inconsistency is still a problem for CI pipelines that use `list_unused_filters: true`.

fyi: Claude Code generated this comment

Ch4s3:
@oliver-kriska I didn't do this initially because it required a lot of file handling. I could possibly do it in the future but I'm waiting on @jeremyjh to take another look at the PR. [This has been my approach in the mean time](https://github.com/jeremyjh/dialyxir/pull/575#issuecomment-3936756154)

epinault:
@jeremyjh any chance we can move this forward? large dialyzer project are slow and while it s harder to justify dialyzer with type safety coming, it would be nice to still support this

Ch4s3:
@epinault I built this to work on a large umbrella and it's used on a few other large apps https://github.com/Ch4s3/assay

epinault:
Is that a replacement then for dialyxir then?

Ch4s3:
@epinault it is for my use case. It's mostly a dumb wrapper around `iplt` generation, checking, and a formatter. Its a little rough, because I'm still interested in merging this PR to dialyxir. If you have any issues trying it, file an issue over there.

test/warning_test.exs
```diff
@@ -4,7 +4,11 @@ defmodule Dialyxir.Test.WarningTest do
   # Don't test output in here, just that it can succeed.
 
   test "pattern match warning succeeds on valid input" do
-    arguments = [~c"pattern {'ok', Vuser@1}", ~c"{'error',<<_:64,_:_*8>>}"]
-    assert(Dialyxir.Warnings.PatternMatch.format_long(arguments))
+    # Avoid exercising Erlex.pretty_print_type/1 paths due to an upstream bug where
```
jeremyjh, line 7: this was resolved upstream

jeremyjh:
I apologize for neglecting this for so long. I was not satisfied with this but not able to put the time in to really understand if it is justified to use a completely bespoke mechanism for app resolution; I think a lot has to change still.

@ypconstante issue was caused by not using Elixir's app resolution mechanism. #596 integrates incremental within the existing framework much more cleanly. I think I prefer that.

Ch4s3:
closing in favor of #602

PR #596, opened 2026-04-03
feat: Add OTP 27 incremental Dialyzer mode support

britton-jb:
Dialyzer's incremental mode (analysis_type: :incremental) is enabled by default on OTP 27+, skipping manual PLT management. Can be overridden via `dialyzer: [incremental: true/false]` in mix config or the --incremental/--no-incremental CLI flags.

### DIFF
--- a/lib/dialyxir/dialyzer.ex
+++ b/lib/dialyxir/dialyzer.ex
@@ -11,7 +11,8 @@ defmodule Dialyxir.Dialyzer do
       :format,
       :list_unused_filters,
       :ignore_exit_status,
-      :quiet_with_result
+      :quiet_with_result,
+      :project_files
     ]
 
     @default_formatter Dialyxir.Formatter.Dialyxir
@@ -42,6 +43,8 @@ defmodule Dialyxir.Dialyzer do
         |> info
 
         {duration_us, result} = :timer.tc(&:dialyzer.run/1, [args])
+
+        result = Dialyxir.Dialyzer.filter_to_project_files(result, split[:project_files])
 
         formatted_time_elapsed = Formatter.formatted_time(duration_us)
 
@@ -86,6 +89,16 @@ defmodule Dialyxir.Dialyzer do
 
       @default_formatter
     end
+  end
+
+  def filter_to_project_files(result, nil), do: result
+
+  def filter_to_project_files(result, project_files) do
+    project_file_set = MapSet.new(project_files)
+
+    Enum.filter(result, fn {_tag, {file, _line}, _warning} ->
+      MapSet.member?(project_file_set, file)
+    end)
   end
 
   @success_return_code 0
--- a/lib/mix/tasks/dialyzer.ex
+++ b/lib/mix/tasks/dialyzer.ex
@@ -27,6 +27,8 @@ defmodule Mix.Tasks.Dialyzer do
       * `--format ignore_file_strict` - format the warnings in {file, short_description} format for Elixir Format ignore file.
     * `--quiet` - suppress all informational messages
     * `--quiet-with-result` - suppress all informational messages except for the final result message
+    * `--incremental` - use Dialyzer's incremental mode (requires OTP 27+, enabled by default on OTP 27+)
+    * `--no-incremental` - disable incremental mode even on OTP 27+
 
   Warning flags passed to this task are passed on to `:dialyzer` - e.g.
 
@@ -148,6 +150,7 @@ defmodule Mix.Tasks.Dialyzer do
   @command_options Keyword.merge(@old_options,
                      force_check: :boolean,
                      ignore_exit_status: :boolean,
+                     incremental: :boolean,
                      list_unused_filters: :boolean,
                      no_check: :boolean,
                      no_compile: :boolean,
@@ -166,17 +169,21 @@ defmodule Mix.Tasks.Dialyzer do
     check_dialyzer()
     compatibility_notice()
 
+    incremental? = incremental_mode?(opts)
+
     if Mix.Project.get() do
       Project.check_config()
 
       unless opts[:no_compile], do: Mix.Task.run("compile")
 
-      _ =
-        unless no_check?(opts) do
-          info("Finding suitable PLTs")
-          force_check? = Keyword.get(opts, :force_check, false)
-          check_plt(force_check?)
-        end
+      unless incremental? do
+        _ =
+          unless no_check?(opts) do
+            info("Finding suitable PLTs")
+            force_check? = Keyword.get(opts, :force_check, false)
+            check_plt(force_check?)
+          end
+      end
 
       default = Dialyxir.Project.default_ignore_warnings()
       ignore_warnings = Dialyxir.Project.dialyzer_ignore_warnings()
@@ -210,7 +217,7 @@ defmodule Mix.Tasks.Dialyzer do
       end
 
       warn_old_options(opts)
-      unless opts[:plt], do: run_dialyzer(opts, dargs)
+      unless opts[:plt], do: run_dialyzer(opts, dargs, incremental?)
     else
       info("No mix project found - checking core PLTs...")
       Project.plts_list([], false) |> Plt.check()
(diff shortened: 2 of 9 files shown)


Ch4s3:
You may have missed this PR https://github.com/jeremyjh/dialyxir/pull/575 and discussion

britton-jb:
Definitely did. Appreciate it @Ch4s3

Ch4s3:
@britton-jb you may want to look at my PR before closing yours, you may have a superior approach. I ended up doing something different with the other library I wrote.

jeremyjh:
@britton-jb This is a much smaller, and more integrated (into existing code paths) implementation than #575. @Ch4s3 anything you see that this is missing from your branch? 

I think most of your doc updates that explain the feature, CI changes, etc would need to come over. Anything else?

Ch4s3:
It's certainly less code. I'm taking a look @jeremyjh. I think the `core_apps` list is worth more consideration, and its useful to think about what the user is expected to declare and document that. If I'm not mistaken this PR doesn't automatically detect the apps the user has in their tree like you requested [here](https://github.com/jeremyjh/dialyxir/pull/575#discussion_r2610557415) which is the source of a good chunk of my code. 

`apps_direct` could probably be removed from my PR and isn't present here which is good, since IIRC you deprecated that.

This PR doesn't seem to support `:warning_apps` which was my way to do the inverse of `plt_ignore_apps` which isn't available with incremental_plts.

This PR probably works, but there's a lot in mine geared towards working well for CI use cases and on large umbrella apps, which IMO are the biggest winners for incremental dialyzer.

I'm happy to try to add commits of my own to this PR, rework mine, or weigh in further.

jeremyjh:
> If I'm not mistaken this PR doesn't automatically detect the apps the user has in their tree like you requested [here](https://github.com/jeremyjh/dialyxir/pull/575#discussion_r2610557415) which is the source of a good chunk of my code.

@Ch4s3 It doesn't because this is already how Dialyxir works; the default resolution mode is `:apps_tree` which does this. https://github.com/jeremyjh/dialyxir

Ch4s3:
@jeremyjh when I did my first pass at this I tired to lean on the existing `:app_tree` with the branch arm that hits [load_external_deps/2](https://github.com/jeremyjh/dialyxir/blob/4a2b6d778fa5fe332ec33b92e6f000bc22b25b7f/lib/dialyxir/project.ex#L389) it put out a lot of noisy errors in incremental mode. 

This because incremental dialyzer has a concept of [warning_files](https://github.com/erlang/otp/blob/5cf5f9725452f4e1b6a4890e8ff0305d76924b98/lib/dialyzer/src/dialyzer_incremental.erl#L784) and [files_rec](https://github.com/erlang/otp/blob/5cf5f9725452f4e1b6a4890e8ff0305d76924b98/lib/dialyzer/src/dialyzer_incremental.erl#L781). The files_rec is a list of all files recursively in any apps you pass to incremental mode. If you pass in just your apps from the tree plus some erlang/elixir apps  you'll get a bunch of `Unknown function Mod:Fun/Arity` and `Unknown type Mod:Type/Arity` warning noise and less precision in your success typing. If you pass in deps without `warning_files` then you get errors/warnings from the deps which isn't what you want either.

So unless users want noise or lost precision this PR needs a way to extract warning files just from project code. Just passing the existing `app_tree` implementation won't do that.

How can I help get this PR landed?

lib/mix/tasks/dialyzer.ex
```diff
@@ -266,18 +273,38 @@ defmodule Mix.Tasks.Dialyzer do
...
+        incremental_plt = String.to_charlist(Project.plt_file() <> ".incremental")
+        project_files = Project.dialyzer_files()
+
+        [
+          {:analysis_type, :incremental},
+          {:init_plt, incremental_plt},
+          {:output_plt, incremental_plt},
+          {:files, project_files ++ dep_beam_files()},
+          {:warnings, dialyzer_warnings(dargs)},
+          {:project_files, project_files}
```
jeremyjh, line 288: @Ch4s3 I think this is right. `Project.dialyzer_files()` will have all the beams from the project under analysis. Those are the only files we want in warning files. We just find the actual beam files of this project recursively.

lib/mix/tasks/dialyzer.ex
```diff
@@ -378,6 +405,27 @@ defmodule Mix.Tasks.Dialyzer do
     end
   end
 
+  defp dep_beam_files do
+    all_apps = Enum.uniq(Project.core_apps() ++ Project.cons_apps())
```
jeremyjh, line 409: @Ch4s3 This should find everything - `cons_apps` will find all deps according to the resolution mode - we pretty much always want `:app_tree` and hopefully with incremental mode we can always afford it memory-wise. So this is the universe of types we may reference.

lib/mix/tasks/dialyzer.ex
```diff
@@ -266,18 +273,38 @@ defmodule Mix.Tasks.Dialyzer do
...
+    base_args =
+      if incremental? do
+        incremental_plt = String.to_charlist(Project.plt_file() <> ".incremental")
+        project_files = Project.dialyzer_files()
+
+        [
+          {:analysis_type, :incremental},
+          {:init_plt, incremental_plt},
+          {:output_plt, incremental_plt},
+          {:files, project_files ++ dep_beam_files()},
```
jeremyjh, line 286: `files_rec` wants the project files along with all their dependencies, recursively; which we should get with `dep_beam_files`. Its the same app list we already use to build PLTs - we just explode and flatten all the beam files in those apps.

lib/dialyxir/dialyzer.ex
```diff
@@ -88,6 +91,16 @@ defmodule Dialyxir.Dialyzer do
     end
   end
 
+  def filter_to_project_files(result, nil), do: result
+
+  def filter_to_project_files(result, project_files) do
+    project_file_set = MapSet.new(project_files)
+
+    Enum.filter(result, fn {_tag, {file, _line}, _warning} ->
```
jeremyjh, line 99: This can't be right though. `project_files` is a list of `.beam` files. `file` here is a source file, an, `.ex` file. @britton-jb Does this branch actually produce warnings on a project that should have them?

jeremyjh (commented):
@Ch4s3 

>  If you pass in just your apps from the tree plus some erlang/elixir apps you'll get a bunch of Unknown function Mod:Fun/Arity and Unknown type Mod:Type/Arity warning noise and less precision in your success typing.

Do you have a public repo that has that issue? It sounds like at the point you saw that issue, you either had an incomplete tree in files_rec or you passed the whole tree as warning_files.  

I think #596 is close to the right app resolution but I don't see how it could actually be working in terms of reporting valid errors.

lib/dialyxir/dialyzer.ex
britton-jb: I believe it is, but I’ll double check, as I’ve been wrong before, and getting it to compile on a long running consistently green project had been the goal.

Ch4s3:
@jeremyjh 

> Do you have a public repo that has that issue? It sounds like at the point you saw that issue, you either had an incomplete tree in files_rec or you passed the whole tree as warning_files.

I do not have a public repo, the only large umbrella app I have access to is from work. This PR doesn't pass `warning_files`, and neither does the main branch because it wasn't a flag for classic plt analysis.

```elixir
{:analysis_type, :incremental},
{:init_plt, incremental_plt},
{:output_plt, incremental_plt},
{:files, project_files ++ dep_beam_files()},   # ← deps go INTO the analysis set
{:warnings, dialyzer_warnings(dargs)},
```
Without the `warning_files` option or the `--warning_apps` cli flag the warning set for the iplt defaults to the whole analysis set, and this would emit Dialyzer warnings on dependency (and OTP) code, not just the project's own modules.

lib/mix/tasks/dialyzer.ex
Ch4s3: This doesn't work the way it looks like it does. Using the existing machinery from dialyxir like `plt_ignore_apps` will remove file from the analysis set, but nothing tells the iplt code what we actually want warnings for so it spits out everything.

Ch4s3:
@jeremyjh and @ britton-jb to make this work correctly with iplts there needs to be a way to define the warning file set. Essentially something [like this](https://github.com/Ch4s3/assay/blob/c26656e0b8f20666410b63361129081443b03613/lib/assay/runner.ex#L183) from my other repo is required. Again, I'm happy to toss a commit onto this PR to implement it, its a small lift. The only fiddly part is providing an automatic way to discern *exactly this project recursively only*.

I don't think post filtering with `filter_to_project_files` is the correct approach.

Ch4s3:
@jeremyjh I think [this is closer to what you want](https://github.com/jeremyjh/dialyxir/pull/602/changes#diff-5c19459efa80671769235504a19c2db6bf014ec50b0b644fc6c1ee084409592bR95) and preserver @britton-jb 's commits.

lib/mix/tasks/dialyzer.ex
jeremyjh: Ah, so that's why the post filtering thing. Good insight.

jeremyjh:
@Ch4s3 agreed, thanks. That looks like the right direction.

Issue #456, opened 2022-03-29
--format ignore_file not working as intended?


x-ji:
### Environment

* Elixir & Erlang/OTP versions (elixir --version):

```
Erlang/OTP 24 [erts-12.2] [source] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1]
Elixir 1.13.3 (compiled with Erlang/OTP 24)
```

* Which version of Dialyxir are you using? (cat mix.lock | grep dialyxir):

```
  "dialyxir": {:hex, :dialyxir, "1.1.0", "c5aab0d6e71e5522e77beff7ba9e08f8e02bad90dfbeffae60eaf0cb47e29488",
[:mix], [{:erlex, ">= 0.2.6", [hex: :erlex, repo: "hexpm", optional: false]}], "hexpm", "07ea8e49c45f15264ebe6d5b93799d4dd56a44036cf42d0ad9c960bc266c0b9a"},
```

### Current behavior
I'm not sure if I didn't correctly understand how it's supposed to work, it seems to me that when I pass in the option `--format ignore_file`, the output is the same as when I don't pass in any `--format` option, and thus cannot be used for `.dialyzer_ignore.exs`?

I'm looking at the output to `stderr`.

### Expected behavior
The output by `mix dialyzer --format ignore_file` can be used in the ignore file.

x-ji:
Putting the output of `mix dialyzer --format dialyzer` in a `dialyzer.ignore-warnings` file worked. Not sure if the new format would be preferred though.

jeremyjh:
Yeah I think the docs could definitely be improved regarding this - its for the legacy warning format.

x-ji:
So I should actually use `mix dialyzer --format short` for the format corresponding to `.dialyzer_ignore.exs`, and `mix dialyzer --format dialyzer` for the old format (`dialyzer.ignore-warnings`), right?

x-ji:
I made a PR at https://github.com/jeremyjh/dialyxir/pull/457. Hope it makes sense (I removed the mention of `--format ignore_file` from the README).

PR #597, opened 2026-05-10
Clarify ignore file formatter docs

dl-alexandre:
## Summary
- clarify that `--format ignore_file` emits `{file, warning_type}` entries for `.dialyzer_ignore.exs`
- clarify that `--format ignore_file_strict` emits `{file, warning_description}` entries and is the recommended granular formatter
- fix the command option wording from "Elixir Format ignore file" to "Elixir term ignore file"

Refs #456

## Validation
- `mix format --check-formatted lib/mix/tasks/dialyzer.ex`
- `git diff --check`

Note: `mix test test/dialyxir/formatter_test.exs test/mix/tasks/dialyzer_test.exs` was attempted, but the Mix task tests timed out building PLTs locally and then left the cwd inside a fixture, causing cascading fixture path failures unrelated to this docs-only change.


jeremyjh:
Thanks!

PR #602, opened 2026-08-27
OTP 27 incremental mode with project-scoped warnings

Ch4s3:
Builds on @britton-jb's  incremental-mode work in #596, and add  warning scoping to the dialyzer args so taht incremental runs report only your project's warnings — not every dependency.

## Background

On OTP 26+, incremental analysis loads the project **and** its dependencies together so callee types resolve. That means dependency beams are part of the analysis set, and without scoping, Dialyzer would report warnings for dependency code too.

#596 solves this with a post-filter, `filter_to_project_files/2`, comparing each warning's file against the project's beam paths. But Dialyzer reports warnings against **source** paths (e.g. `lib/foo.ex`), not beam paths — so the comparison never matched and *every* warning was dropped. Incremental runs reported a false clean. The unit test passed only because it used beam paths on both sides of the comparison, a shape real Dialyzer output never produces.

## This PR

Replaces the post-filter with Dialyzer's own `warning_files_rec` option (the programmatic form of the `--warning_apps` CLI flag): analyze everything, report only the configured apps.

- **Defaults** to the project's own compiled paths (umbrella-aware).
- **Override** with `dialyzer: [warning_apps: [...]]` config or the new `--warning-apps app_a,app_b` CLI flag.
- Dependencies stay in the analysis set (types resolve, no `:unknown` warnings) but are silent unless opted in.

`filter_to_project_files/2` and its `:project_files` internal arg are removed.


Ch4s3:
@jeremyjh I have 2 questions for you. Should we include OTP26? Do you want a way to manually set the analysis set apps, the code I built on top of does it automatically?


The manual analysis part can be partially covered by adding things to `plt_add_apps`, e.g. :`sasl`, `:mix`, `:eex`, `:ssl`, `:public_key`, but increment dialyzer doesn't/can't use the plt_add_apps flag so it works incidentally because `cons_apps()` covers it. You need to do this if your app directly calls mix, ssl, sasl code or whatever. But I think this is really misleading and precludes cases where you might want a narrow analysis set.

jeremyjh:
@Ch4s3 good questions

> I have 2 questions for you. Should we include OTP26? 

We can drop it from CI. Our CI policy is to include the three most recent major releases for Elixir & Erlang. However, we shouldn't set any lower bounds that we don't actually know are a problem.

I guess your question is do we need to handle the incremental flag with it? And if so do we test it? If we can handle that and test that in a reasonably sound way that doesn't involve actually keeping it in the build matrix (e.g. just a mock of some kind) then I think that would be kind of us. 

It is not required and arguably not worth even a little bit of complexity. So I'm fine with whatever you decide. We could also just document that it doesn't work - in reality the projects still on OTP 26 are not updating their deps and using new features.

>  increment dialyzer doesn't/can't use the plt_add_apps flag so it works incidentally because `cons_apps()` covers it.

I'm not clear if there is actually any gap here?  If cons_apps includes them we send them along and they would be used for code that depends on those types, right? This is why I thought we should use `cons_apps`. We have `:plt_add_apps` mostly because there are some large apps with many dependencies and little memory in CI who have to use `:apps_direct` and then add in some of their apps (and probably ignore some warnings) because they can't build their entire PLT otherwise. I'm actually hopeful incremental might entirely solve that - as I understand it will only include types we actually use.

Ch4s3:
@jeremyjh 
> as I understand it will only include types we actually use.

Incremental mode pulls in every app file int he analysis set. That's why I you need to set the warning files to narrow the returned errors.

jeremyjh:
> Incremental mode pulls in every app file int he analysis set. That's why I you need to set the warning files to narrow the returned errors.

I see, you are right. I thought warning apps actually narrowed the scope of the analysis to the actual types needed from the full set of apps, but based on what the docs say it is just filtering warnings. So probably the initial PLT is going to take just as much time and memory as build a full traditional PLT, or even more since it also has to analyze the project files too.

Ch4s3:
> So probably the initial PLT is going to take just as much time and memory as build a full traditional PLT

Yes, but it's usually within a few seconds in my tests. incremental runs are very fast for most smaller changes though. Do you want a manual way to set just the desired apps?

jeremyjh:
> > So probably the initial PLT is going to take just as much time and memory as build a full traditional PLT
> 
> Yes, but it's usually within a few seconds in my tests. incremental runs are very fast for most smaller changes though. Do you want a manual way to set just the desired apps?

@Ch4s3 I don't think we'll need another one - since we use `Project.cons_apps` that can still be done by setting `plt_add_deps: false, plt_apps: [:thisone, :thatone]`.

Ch4s3:
@jeremyjh I'll try to get this in shape for review tomorrow.

PR #604, opened 2026-09-21
Fix operator precedence for loading apps in plt_add_apps

jessestimpson:
Dialyxir loads apps that are defined in `:plt_apps`, but when `:plt_add_apps` is defined, `load_apps/0` is skipped due to an issue with operator precedence.

This patch ensures that the default `[]` is respected, and when `:plt_add_apps` is defined, `load_apps/0` is called as expected.


jeremyjh (approved):

jeremyjh:
@jessestimpson thanks!

